PR #28, opened 2017-10-31
refactor inclusion of sub-classes

LongLiveCHIEF:
Removing the usage of class-like resource declaration to order sub-classes in the module.

This is the first step required to enable more modern module design patterns.


scotty-c:
LGTM

PR #29, opened 2017-11-01
allow configuration of gem sources

LongLiveCHIEF:
This contribution allows developers to configure the source of developer dependencies. This allows developers behind a firewall to point to an approved mirrored source of the publicly available gems that are used to develop and test the module.


scotty-c:
LGTM

PR #35, opened 2017-11-08
Allow all digits (IP) for proxy URL

ianssoftcom:
In our environment the proxy URL is an IP address, the regex did not allow the last octet to be numeric.


davejrt:

PR #41, opened 2017-11-14
repos.pp: use ensure_packages instead of package resource

hex2a:


scotty-c:
LGTM thanks for the update

PR #52, opened 2017-11-20
Allow multiple docker::secrets declarations

### LINKED ISSUE #51: Can't declare more then one docker::secrets in the same catalog
When declaring more then two `docker::secrets` resources the module will throw an error like this

```
...
Duplicate declaration: Exec[docker secret create]
...
```

This happens because `exec` statement doens't  include `$title` in its name.
I'm managing to fix this and will submit a PR soon

guimaluf:
When declaring two or more docker::secrets, catalog will fail with
/Duplicate declaration: Exec[docker secret create]/

This commit includes ${title} in exec resources so it doens't throw any
error

Fixes #51


davejrt:
LGTM

PR #59, opened 2017-11-24
fixing incorrect repo url for redhat

davejrt:


manifests/init.pp
```diff
@@ -553,7 +553,7 @@
             $release = $docker_ce_release
             }
           'Redhat' : {
-            $package_location = "https://download.docker.com/linux/${os}/${::operatingsystemmajrelease}/${::architecture}/${docker_ce_channel}"
+            $package_location = "https://download.docker.com/linux/centos/${::operatingsystemmajrelease}/${::architecture}/${docker_ce_channel}"
```
MWilsonPuppet: @davejrt Hey Dave it looks like for Centos ${::operatingsystemmajrelease}/${::architecture}/${docker_ce_channel}" are not required. Repo URL looks like it should be $package_location = "https://download.docker.com/linux/centos/"
Cheers
Mark

davejrt: hey @MWilsonPuppet without those you'll end up with the follow error:

```puppet
https://download.docker.com/linux/centos/repodata/repomd.xml: [Errno 14] HTTPS Error 404 - Not Found
```

davejrt: @MWilsonPuppet the full string is required as the repomd.xml is at that URL. Also if we didn't use that then we wouldn't be differentiating between the stable/test/edge channels

MWilsonPuppet: @davejrt Ahhh my bad Dave apologies. Always 5 steps ahead of me :).
cheers
Mark

PR #60, opened 2017-11-27
Additional flags for creating a network

coder-hugo:
This PR adds the functionality to pass additional flags to the `docker network create` command as there are several missing flags in the current implementation of the `docker_network` resource. In my current use case I just need the `--internal` flag. So another solution could be to just add a new parameter `internal` for the `docker_network` resource but I decided to do this with a more generic solution. With this solution it's possible to use all of the currently available flags and all the flags that may be available in the future without adapting this module again.


davejrt:

PR #61, opened 2017-11-27
(CLOUD-1560)-adding-functionality to remove systemd service files.

MWilsonPuppet:
This PR adds functionality so that when docker run is set to ensure => absent,  the systemd service files are removed. I have verified that this works with single and multiple containers and it is working as expected.


davejrt:
@MWilsonPuppet this is looking good so far. Can you also update the unit tests so we have coverage of this?

MWilsonPuppet:
@davejrt Hey Dave I have added the spec and acceptance tests and updated the readme.
Cheers
Mark

davejrt:
LGTM

PR #63, opened 2017-11-28
Allow skipping usage of pw_hash

### LINKED ISSUE #62: docker::registry: Allow password to be provided as a hash
`pw_hash` has issues with FIPS systems ([MODULES-6135](https://tickets.puppetlabs.com/browse/MODULES-6135)), so I would like to be able to provide a password as a hash to skip the `pw_hash` function. 

The line https://github.com/puppetlabs/puppetlabs-docker/blob/master/manifests/registry.pp#L71, which specifies SHA-512, still fails on `pw_hash` (weird). 

My test environment can be replicated by cloning [simp/simp_docker](https://github.com/simp/pupmod-simp-simp_docker), installing the gems etc, and running `BEAKER_fips=yes BEAKER_destroy=no rake 'beaker:suites[redhat]'`. It should fail on this line: https://github.com/simp/pupmod-simp-simp_docker/blob/master/spec/acceptance/suites/redhat/30_registry_spec.rb#L81

jeefberkey:
Added a param to docker::registry to allow a user to skip using puppet
to generate the password hash, instead specifying it themselves. This
will allow this module to work on FIPS systems.

Closes #62

### DIFF
--- a/manifests/registry.pp
+++ b/manifests/registry.pp
@@ -18,6 +18,9 @@
 #   Password for authentication to private Docker registry. Leave undef if
 #   auth is not required.
 #
+# [*pass_hash*]
+#   The hash to be used for receipt. If left as undef, a hash will be generated
+#
 # [*email*]
 #   Email for registration to private Docker registry. Leave undef if
 #   auth is not required.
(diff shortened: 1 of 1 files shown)


scotty-c:
LGTM, thanks for your contribution.

PR #69, opened 2017-12-06
update docs for using legacy package

alexcreek:
If you don't disable the service_overrides_template when using the legacy packages, you get this horrible error when a container starts:
``` 
/usr/bin/docker-current: Error response from daemon: shim error: docker-runc not installed on system.
```
Adding what worked for me to the docs ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,12 +94,13 @@ class { 'docker':
 ```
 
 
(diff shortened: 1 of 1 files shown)


scotty-c:
LGTM

PR #71, opened 2017-12-06
Add support for overlay2.override_kernel_check setting

gdubicki:
Quoting https://docs.docker.com/engine/reference/commandline/dockerd/ :

> overlay2.override_kernel_check
> 
> Overrides the Linux kernel version check allowing overlay2. Support for specifying multiple lower directories needed by overlay2 was added to the Linux kernel in 4.0.0. However, some older kernel versions may be patched to add multiple lower directory support for OverlayFS. This option should only be used after verifying this support exists in the kernel. Applying this option on a kernel without this support will cause failures on mount.

This is required to use overlay2 on Centos 7 with kernel 3.x.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -279,6 +279,10 @@
 #   synchronization, to continue even though the configuration may be buggy.
 #   Defaults to true
 #
+# [*overlay2_override_kernel_check*]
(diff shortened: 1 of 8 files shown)


gdubicki:
Tested on Centos 7.4, it works.

davejrt:
LGTM

PR #76, opened 2017-12-11
(maint)Fixing syntax error to get ci passing

MWilsonPuppet:
This PR fixes a syntax error that puppet-lint-unquoted_string-check version 0.2.5 was complaining about.CI uses the internal ruby Gems and so is not using the latest it uses 0.2.5 which is why we didn't see this error when running locally as we use latest locally (0.3).

 lt also fixes a deprecation warning for the deprecated metadata task and replaces with metadata_lint. It also removes spec tests from running during test rake task so that  test now runs lint and metadata tests. This is so the spec tests are not running twice in ci. Verified that all tests still passing.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -54,7 +54,7 @@ begin
     :syntax,
     :lint,
     :parallel_spec,
-    :metadata,
+    :metadata_lint,
   ]
 rescue LoadError # rubocop:disable Lint/HandleExceptions
 end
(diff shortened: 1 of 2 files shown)


MWilsonPuppet:
@davejrt Hey Dave for some reason this isnt letting me add you as a request to review. Could you review this if you have a minute.
Cheers
Mark

scotty-c:
LGTM

Issue #84, opened 2017-12-19
swarm legacy (docker 17.05) package regex doesn't match package pattern for centos


dmcanally:
The package available from docker for centos 7 is...

17.05.0.ce-1.el7.centos

due to this, the regex at line 540 in init.pp needs to be updated...

https://github.com/puppetlabs/puppetlabs-docker/blob/d04b90dd00d92f89136bbb182b8295ddc18a56dc/manifests/init.pp#L540

`  if ( $version == undef ) or ( $version !~ /^(17[.]0[0-5][.]\d(~|-)ce|1.\d+)/ ) {`
I believe should be...
`  if ( $version == undef ) or ( $version !~ /^(17[.]0[0-5][.]\d(~|-|\.)ce|1.\d+)/ ) {`
I'll submit a pull request shortly w/ the change.

davejrt:
Thanks for your PR @dmcanally

PR #85, opened 2017-12-19
Adding \. to regex on init.pp:540 to address issue #84

dmcanally:
this is to address issue #84


davejrt:

PR #86, opened 2017-12-20
docker::run: Add service_provider param to override OS detection

iamjamestl:
I know support for Gentoo in this module was dropped when it moved into the puppetlabs space.  That said, it is still the easiest way to get Docker up and running on Gentoo with just a few extra parameters:

```puppet
class { 'docker':
  docker_ce_package_name     => 'app-emulation/docker',
  service_config             => '/etc/conf.d/docker',
  service_config_template    => 'docker/etc/sysconfig/docker.systemd.erb',
  service_overrides_template => 'mine/docker/systemd-overrides.conf.erb',
  service_provider           => 'systemd',
}
```

where the `mine/docker/systemd-overrides.conf.erb` template contains:

```ini
[Service]
EnvironmentFile=-/etc/conf.d/docker
ExecStart=
ExecStart=/usr/bin/dockerd $OPTIONS
```

Unfortunately, the `docker::run` type detects that it is running on Gentoo and sets things up for OpenRC.  Gentoo has first class support for systemd and I'd like to be able to tell the run type to use `systemd`, just as I do with `Class[docker]` via the `service_provider` parameter.

This pull request adds a `service_provider` parameter to the `docker::run` type which can be set to override the operating system detection.  It can be set to `systemd` to enable systemd support.  If it is set to anything else, or it is unset (the default), then the type falls back to the OS detection behavior that was there before.

The default behavior of the type is unchanged.

I have added rspec tests to validate this behavior.

This pull request also contains a couple of trivial variable definitions in default cases that were left undefined on non-RedHat and Debian systems.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -563,7 +563,11 @@ class docker(
             $package_key_source = $docker_ce_key_source
             $package_key_check_source = true
           }
-          default: {}
+          default: {
+            $package_location = $docker_package_location
+            $package_key_source = $docker_package_key_source
+            $package_key_check_source = $docker_package_key_check_source
+          }
         }
         $docker_start_command = $docker_ce_start_command
         $docker_package_name = $docker_ce_package_name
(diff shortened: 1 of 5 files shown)


davejrt:
There is some planned work in the pipeline the remove a lot of the existing logic referring to other OS. I'd like to have that work completed before we add additional functionality now that may create more tech debt. 

When the ticket makes it into the sprint I'll make sure the work you've done here is taken into consideration. Thanks for your contribution. Feel free to grab me on the puppet community slack if you want to discuss further.

Issue #87, opened 2017-12-20
unable to remove docker services


RXM307:
as per the README I tried adding
    docker::services {'filebeat':
      ensure => 'absent',
      service_name => 'filebeat',
    }

but as it defaults to create = true in services.pp so I tried create => false but it still tries to create

    docker::services {'filebeat':
      create => false,
      ensure => 'absent',
      service_name => 'filebeat',
    }

Exec[Docker service create]/returns: "docker service create" requires at least 1 argument(s).
Exec[Docker service create]/returns: See 'docker service create --help'.
Exec[Docker service create]/returns:
Exec[Docker service create]/returns: Usage:  docker service create [OPTIONS] IMAGE [COMMAND] [ARG...]
Exec[Docker service create]/returns:
Exec[Docker service create]/returns: Create a new service
Error: docker service create --name 'filebeat' returned 1 instead of one of [0]
Exec[Docker service create]/returns: change from notrun to 0 failed: docker service create --name 'filebeat' returned 1 instead of one of [0]

should services.pp be set to
if $create == 'true' {
instead of
if $create {

and the same for
if $update {
and
if $scale {

RXM307:
after further testing it looks like you must specify

  create => false,

I accidentally had the docker::services specified twice in my manifest and it was only hitting the first occurrence.

if this is by design should the README be updated?

scotty-c:
Hi, @RXM307 this is by design, I have updated the README. Thanks for letting us know.

PR #89, opened 2017-12-21
 Fix for Issue #87 and remove docker_command hardcoding

RXM307:


RXM307:
Looks like this is failing CI Build, I think the original issue was from an older build with synxtax errors

PR #90, opened 2017-12-22
Allow setting commented TMPDIR in systemd file for backwards compatibility

esalberg:
The tmpdir fix in #58 was not backwards compatible, in that there was no way to retain the legacy config (so that we can restart docker for some nodes sooner than others, which happens when that file is updated).

Added a flag tmp_dir_config defaulting to true, which sets TMPDIR in the config.  Setting it to false has TMPDIR commented out (legacy behavior).


davejrt:
LGTM, thanks for your PR

PR #99, opened 2018-01-04
(CLOUD_1505) validate functions

sheenaajay:
Request to review the changes for replacing validate functions with datatypes and assert_type function.
Following jobs are running clean
bundle exec rake test
bundle exec rake spec
bundle exec rake acceptance

### DIFF
--- a/manifests/compose.pp
+++ b/manifests/compose.pp
@@ -21,14 +21,12 @@
 #   Proxy to use for downloading Docker Compose.
 #
 class docker::compose(
-  $ensure = 'present',
-  $version = $docker::params::compose_version,
-  $install_path = $docker::params::compose_install_path,
-  $proxy = undef
+  Optional[Pattern[/^present$|^absent$/]] $ensure          = 'present',
+  Optional[String] $version                                = $docker::params::compose_version,
+  Optional[String] $install_path                           = $docker::params::compose_install_path,
+  Optional[String] $proxy                                  = undef
 ) inherits docker::params {
-  validate_string($version)
-  validate_re($ensure, '^(present|absent)$')
-  validate_absolute_path($install_path)
+
   if $proxy != undef {
       validate_re($proxy, '^(https?:\/\/)?([^:^@]+:[^:^@]+@|)([\da-z\.-]+)\.([\da-z\.]{2,6})(:[\d])?([\/\w \.-]*)*\/?$')
   }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -353,163 +353,157 @@
 #    Sets the tmp dir for Docker (path)
 #
 class docker(
-  $version                           = $docker::params::version,
-  $ensure                            = $docker::params::ensure,
-  $prerequired_packages              = $docker::params::prerequired_packages,
-  $docker_ce_start_command           = $docker::params::docker_ce_start_command,
-  $docker_ce_package_name            = $docker::params::docker_ce_package_name,
-  $docker_ce_source_location         = $docker::params::package_ce_source_location,
-  $docker_ce_key_source              = $docker::params::package_ce_key_source,
-  $docker_ce_key_id                  = $docker::params::package_ce_key_id,
-  $docker_ce_release                 = $docker::params::package_ce_release,
-  $docker_package_location           = $docker::params::package_source_location,
-  $docker_package_key_source         = $docker::params::package_key_source,
-  $docker_package_key_check_source   = $docker::params::package_key_check_source,
-  $docker_package_key_id             = $docker::params::package_key_id,
-  $docker_package_release            = $docker::params::package_release,
-  $docker_engine_start_command       = $docker::params::docker_engine_start_command,
-  $docker_engine_package_name        = $docker::params::docker_engine_package_name,
-  $docker_ce_channel                 = $docker::params::docker_ce_channel,
-  $docker_ee                         = $docker::params::docker_ee,
-  $docker_ee_package_name            = $docker::params::package_ee_package_name,
-  $docker_ee_source_location         = $docker::params::package_ee_source_location,
-  $docker_ee_key_source              = $docker::params::package_ee_key_source,
-  $docker_ee_key_id                  = $docker::params::package_ee_key_id,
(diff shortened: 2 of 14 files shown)


manifests/registry.pp
```diff
@@ -73,7 +73,7 @@
     $local_user_strip = regsubst($local_user, '-', '', 'G')
 
     $_pass_hash = $pass_hash ? {
-      Undef   => pw_hash("${title}${auth_environment}${auth_cmd}${local_user}", 'SHA-512', $local_user_strip),
+      'Undef'   => pw_hash("${title}${auth_environment}${auth_cmd}${local_user}", 'SHA-512', $local_user_strip),
```
davejrt: @sheenaajay have you run this through travis without the qoutes? The PR to remove them was valid and needs to as is.

manifests/init.pp
```diff
@@ -357,163 +357,184 @@
...
-  validate_array($daemon_environment_files)
-  validate_array($log_opt)
-  validate_bool($tls_enable)
-  validate_bool($ip_forward)
-  validate_bool($iptables)
-  validate_bool($ip_masq)
-  if $icc != undef {
-    validate_bool($icc)
+  if $version {
+    assert_type(String[1], $version)
```
HelenCampbell: Not 100% sure, but I think you can put this check in the type declaration itself, like - Optional[String[1]] $version - Will save you the extra couple of lines of the assert type here :)

manifests/init.pp
```diff
@@ -357,163 +357,184 @@
...
-  validate_string($fixed_cidr)
-  validate_string($default_gateway)
-  validate_string($bip)
 
+  if $::osfamily {
+    assert_type(Pattern[/^(Debian|RedHat|Archlinux|Gentoo)$/], $::osfamily) |$a, $b| {
+      fail('This module only works on Debian or Red Hat based systems or on Archlinux as on Gentoo.')
+    }
+  }
+  assert_type(Boolean, $manage_kernel)
```
HelenCampbell: When you define the params at the top with the Data Type it's pretty much the same as doing an assert type on the param - If it comes through as anything other than the specified Data Type it'd fail further up at the param definition, so you can take all these assert types out!

manifests/init.pp
```diff
@@ -541,15 +562,19 @@
     if(!$tcp_bind) {
         fail('You need to provide tcp bind parameter for TLS.')
     }
-    validate_string($tls_cacert)
-    validate_string($tls_cert)
-    validate_string($tls_key)
+    assert_type(String[1], $tls_cacert)
```
HelenCampbell: Again, if it makes sense, I'd remove these assert types and add the string lower limit to the data type - Optional[String[1]] $tls_cacert

manifests/registry.pp
```diff
@@ -33,19 +33,19 @@
...
+  Optional[String] $pass_hash   = undef,
+  Optional[String] $email       = undef,
+  Optional[String] $local_user  = 'root',
+  Optional[String] $version     = $docker::version,
+  Optional[Boolean] $receipt     = true,
 ) {
   include docker::params
 
-  validate_re($ensure, '^(present|absent)$')
+  assert_type(Pattern[/^present$|^absent$/], $ensure)
```
HelenCampbell: You can use Pattern[] as a data type! So you could remove this line and have the definition further up as -
 Optional[Pattern[/^present$|^absent$/]] $ensure - https://puppet.com/docs/puppet/5.3/lang_data_abstract.html#pattern

manifests/run.pp
```diff
@@ -120,39 +120,53 @@
   $service_name = $docker::service_name
   $docker_group = $docker::docker_group
 
-  validate_re($image, '^[\S]*$')
-  validate_re($title, '^[\S]*$')
-  validate_re($memory_limit, '^[\d]*(b|k|m|g)$')
-  validate_re($ensure, '^(present|absent)')
+  if $image {
+  assert_type(Pattern[/^[\S]*$/], $image)
```
HelenCampbell: As above - I'd consider moving the params here to data type Pattern to be able to remove the assert_types here.

sheenaajay:
@davejrt Thanks alot Dave.Will check the local environment and reexecute the unit testcases.Travis jobs are running clean.

sheenaajay:
@HelenCampbell Thanks alot for the review.

manifests/compose.pp
```diff
@@ -21,17 +21,11 @@
...
 #
 class docker::compose(
-  $ensure = 'present',
-  $version = $docker::params::compose_version,
-  $install_path = $docker::params::compose_install_path,
-  $proxy = undef
+  Optional[Pattern[/^present$|^absent$/]] $ensure                                                                         = 'present',
+  Optional[String] $version                                                                                               = $docker::params::compose_version,
+  Optional[String] $install_path                                                                                          = $docker::params::compose_install_path,
+  Optional[Pattern[/^(https?:\/\/)?([^:^@]+:[^:^@]+@|)([\da-z\.-]+)\.([\da-z\.]{2,6})(:[\d])?([\/\w \.-]*)*\/?$/]] $proxy = undef
```
jeefberkey: I have to scroll to see what the parameter names are, maybe a custom type alias is a good idea here

sheenaajay:
@davejrt @HelenCampbell Thanks for the reviews.All comments are incorporated.

manifests/compose.pp
jeefberkey: Same comment for every large data type

sheenaajay: @jeefberkey Thank you for the valuable comment

sheenaajay: @jeefberkey The large data types are not repeating alot to create a data type.So will specify the assert_type function for all large data types outside the variable declaration area so keep it short.

davejrt:
LGTM

esalberg:
We are having issues upgrading past this commit as well due to dm_use_deferred_deletion and dm_use_deferred_removal being set to String instead of Boolean.  Perhaps it was set to String because the default was set to 'undef' (converted to false)?  My concern is that while the String update breaks our config, I don't know if other installations might be affected the other way (although lint doesn't like a quoted true/false, so I'm assuming most probably used booleans by default).

PR #103, opened 2018-01-07
Add facts from docker version and docker info.

mterzo:
Adding version, info and swarm tokens to facts when available.


mterzo:
I noticed that travis is not running "rake spec".

mterzo:
Is there a suggestion for distributing the tokens through the cluster then?

davejrt:
We recommend using puppet tasks

mterzo:
Backed out tokens.

davejrt:
LGTM

PR #109, opened 2018-01-10
CLOUD-1610 Adding registry mirror flag

MWilsonPuppet:
This PR adds the registry-mirror flag so that the user can specify a preferred registry mirror. I have verified that it is working as expected. After the manifest is applied docker system info shows the specified registry mirror. The readme has been updated and unit and acceptance tests added. Unit and acceptance tests are passing.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -352,6 +352,10 @@
 # [*tmp_dir*]
 #    Sets the tmp dir for Docker (path)
 #
+# [*registry_mirror*]
+#   Sets the prefered container registry mirror.
+#   Default: undef
+#    
 class docker(
   Optional[String] $version                           = $docker::params::version,
   String $ensure                                      = $docker::params::ensure,
@@ -462,7 +466,7 @@ class docker(
   Optional[String] $service_overrides_template        = $docker::params::service_overrides_template,
   Optional[Boolean] $service_hasstatus                = $docker::params::service_hasstatus,
   Optional[Boolean] $service_hasrestart               = $docker::params::service_hasrestart,
-
+  Optional[String] $registry_mirror                   = $docker::params::registry_mirror,
 ) inherits docker::params {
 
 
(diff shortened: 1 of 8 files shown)


manifests/init.pp
```diff
@@ -462,7 +466,7 @@
   Optional[String] $service_overrides_template        = $docker::params::service_overrides_template,
   Optional[Boolean] $service_hasstatus                = $docker::params::service_hasstatus,
   Optional[Boolean] $service_hasrestart               = $docker::params::service_hasrestart,
-
+  Optional[String] $registry_mirror                   = $docker::params::service::registry_mirror,
```
davejrt: @MWilsonPuppet can you review this line, it's failing the unit tests

MWilsonPuppet: @davejrt Hey Dave I dug in and  issue was with  login in ERB template. I have made changes to the erb template and tested working also added a negative test case. Ill push  changes now. Just need to add a final change to read me .

templates/etc/sysconfig/docker.systemd.erb
```diff
@@ -9,6 +9,7 @@ OPTIONS="<% if @root_dir %> -g <%= @root_dir %><% end -%>
  --iptables=<%= @iptables -%>
  --ip-masq=<%= @ip_masq -%>
 <% unless @icc == nil %> --icc=<%= @icc %><% end -%>
+<% if @registry_mirror != 'undef' %> --registry-mirror=<%= @registry_mirror %><% end -%>
```
davejrt: @MWilsonPuppet we can use the same pattern as the other params here. We only need to test if registry_mirror is defined, not that it's contents doesn't equal the string 'undef'. 

so I'm thinking something like this:

```
<% if @registry_mirror %> --registry-mirror=<%= @registry_mirror %><% end -%>
```

Then change the default value of registry mirror, back to undef, not 'undef'

MWilsonPuppet: @davejrt Ive made the requested changes and  acceptance and unit tests are passing.
cheers
Mark

davejrt:
LGTM

PR #111, opened 2018-01-11
Add resource type and provider for docker_volumes

dhollinger:
Added a native resource for managing Docker volumes. There is a new
type and provider (docker_volume) and a docker::volumes class for passing
hieradata to the resource.

RSpec and Beaker tests have been added and the docs have been updated
with information on the new resource.


davejrt:

PR #115, opened 2018-01-15
Repo fix arch

davejrt:


RussellMcOrmond:
Should I assume the error I got when I tried to apply this patch is because I'm running an older version of puppet?

2018-01-15 10:32:58,007 ERROR [puppet-server] Puppet facts is not a hash or array when accessing it with architecture at /etc/puppet/environments/production/modules/docker/manifests/repos.pp:8 on node quark.office.c7a.ca

```
root@kernhem:~# puppetserver --version
puppetserver version: 1.2.0
root@kernhem:~# puppet --version
3.8.7
root@kernhem:~# 

```

davejrt:
Correct

The way this fact is referenced is supported in Puppet 4 onwards which is
our minimum support base for this module.

PR #116, opened 2018-01-15
removing manage_kernel

davejrt:
This PR removes the manage_kernel param as it's no longer required for the supported OS versions in this module


scotty-c:
LGTM

PR #121, opened 2018-01-18
replace Tuple with Array[String]

lobeck:
if an array is passed to a Variant[Tuple] it causes `Error: Puppet::Parser::Compiler failed with error NoMethodError: undefined method `instance?' for nil:NilClass on node`

This is probably due to the wrong use of Tuple, as the type parameter is mandatory. As we want to support multiple values, we need to use Array[String]


davejrt:

PR #128, opened 2018-01-25
Cloud 1528 removing unsupported operating systems from the module 

MWilsonPuppet:
This PR Deprecates  Gentoo , Archlinux, Debian 7, Centos 6 and Ubuntu 12 .04 so that the supported OS versions are correct. The spec and acceptance tests have also been updated changing from Debian 7 to 9 and Centos 6 to 7. Spec and acceptance tests are passing.


davejrt:
Closing as this work has been includined in PR #129

jantman:
Is there an alternate module that adds back some of these OSes - specifically Arch Linux? I've been happily using garethr/docker on Arch for years, until systemd 241 broke it within the past few weeks. garethr/docker suggests this module as a replacement, but it seems this no longer supports Arch linux?

PR #146, opened 2018-02-05
Add more supported params to run example

lionax:
Some more of the supported params. Some are still missing where I don't know how they work.


davejrt:
LGTM

PR #154, opened 2018-02-09
Add new parameter 'read_only' to ::docker::run

### LINKED ISSUE #153: Native read-only parameter for docker run
Quite often, I'm finding myself having to pass `--read-only` as an `extra_parameter` to `::docker::run`, like so:
```puppet
 ::docker::run { 'postgres':
    image            => "puppet/puppetdb-postgres:${version}",
    ...
    extra_parameters => [
      '--read-only=true',
    ]
}
```
It would be great to have a 'native' parameter for this, i.e. `read_only => true`.

jacksgt:
Fixes #153


davejrt:

PR #157, opened 2018-02-13
Fixing typo that breaks docker-compose file version matching and enab…

### LINKED ISSUE #156: Docker compose version is not recognized correctly and acceptance tests for compose v2 are disabled
I noticed a strange error: 

> Could not evaluate: Unsupported docker compose file syntax version "2"!

Diving a little deeper i found the issue, [a bad regex](https://github.com/puppetlabs/puppetlabs-docker/blob/master/lib/puppet/provider/docker_compose/ruby.rb#L19)

Then I wondered why it passed acceptance testing and turns out they were not run because of bad naming they should be named **compose_v2_spec.rb** instead of **compose_spec_v2.rb**

Fixing the test naming makes the issue show up in the testing.

sigsegv11:
Should resolve #156


davejrt:
LGTM

PR #160, opened 2018-02-16
Add registry_mirror parameter to services to match services flag function

esalberg:
The docker_service_flags function had a option added for registry-mirror without a matching parameter in the services manifest.  (Docker threw an error configuring a service because --registry-mirror was set without a value.)


davejrt:
LGTM

PR #162, opened 2018-02-19
Use `-o` option when downloading docker compose

### LINKED ISSUE #159: docker::compose creates 0 byte /usr/local/bin/docker-compose-1.9.0
If https://github.com/puppetlabs/puppetlabs-docker/blob/9ac570b3a1a3cf1a5f45a8a844b31f3413761401/manifests/compose.pp#L46 fails, it creates a 0 byte file and the download is not reattempted on the next puppet run.

Instead of redirecting stdout to a file, better to use curl's `-o` option?

alexjfisher:
Fixes #159

Also use the `-S` option.  From curl's man page.
> -S, --show-error
>   When used with -s it makes curl show an error message if it fails.


scotty-c:
LGTM

PR #169, opened 2018-02-21
(FM-6817) Updated to the latest version of docker-compose

genebean:
The previous version does not support v3 of docker-compose.yml... this
one does.


scotty-c:
@genebean is this PR really needed as the user can specify the version they need. If we update this and there are any users that have not specified a version Puppet will update there compose version without them knowing

genebean:
@scotty-c Docker stacks (the way you do compose when in swarm mode) say they require v3 compose files... since the module supports stacks it seems logical to me that it should default to a version of compose that is compatible.

genebean:
Can we at least get this documented then? The parameter in question is not in the readme at all (unless I missed it) and I did not know this was a setting until digging into the code. An example that shows how to use a newer version and documents the default version and its limitations would be nice.

scotty-c:
@genebean we have updated the README https://github.com/puppetlabs/puppetlabs-docker/blob/master/README.md#compose

genebean:
@scotty-c Thanks for updating it!

PR #171, opened 2018-02-22
Added option to pass a Dockerfile to $docker_dir

genebean:
This was needed because some images would not build when only a
directory or only a dockerfile were specified. In paticular, this issue
cropped up when building from source and using a dockerfile with an
alternate name.


davejrt:
LGTM

PR #177, opened 2018-03-02
Fix debian systemd

Damoun:
Hi,

This PR fix `docker::run` on debian with systemd.
The container need to not be detached when running systemd.

I removed a unused variable, some duplicate code and replaced a few variables by using `docker::params` instead.

Let me know if you have any question/review on this PR.


davejrt:
LGTM

PR #178, opened 2018-03-11
[WIP] Add Archlinux compatibility

bastelfreak:


davejrt:
@bastelfreak we won't be supporting Archlinux with this module. It was removed specifically, along with Gentoo.

bastelfreak:
Could you explain why it was removed? There are still parts in the code. What is the issue with this specific module? The modules team accepted those PRs for years on other modules.

davejrt:
When the module was forked into the puppetlabs namespace and updated to support docker-ce/ee. The decision was made then to no longer support archlinux and gentoo as they are not officially supported upstream. 

Which code are you referring to that specifically reference support for archlinux?

scotty-c:
@bastelfreak We are only supporting OS's that Docker officially supports in this module. As the Archlinux package for Docker does not come from Docker we cant know its quality. Also, we are only supporting OS  types Puppet officially supports  https://puppet.com/docs/puppet/5.0/system_requirements.html#puppet-agent-and-operating-system-support-life-cycles. So we will not be accepting this PR.

hunner:
> Also, we are only supporting OS types Puppet officially supports https://puppet.com/docs/puppet/5.0/system_requirements.html#puppet-agent-and-operating-system-support-life-cycles.

@davejrt @scotty-c Compatibility and support are often used interchangeably and can lead to confusion. Just for reference, "the modules team's" official stance on this topic (puppetlabs-docker is not maintained by "the modules team" though) is that **support** is only offered officially on supported modules for OSs specified in that link that are also in the `metadata.json` compatibility field. Other operating systems may be **compatible** with various module/OS combinations (such as stdlib/mysql/apache or freebsd/archlinux/fedora) by community effort, but no testing on our part exists to verify or support that compatibility.

We don't generally turn away quality pull requests to add compatibility, only PRs that try to add support claims, or ones that require otherwise-unneeded extensive rewrites to be compatible. It would be much preferred to have more contributors to our codebases via unsupported compatibility that pushing users to fork and compete.

@bastelfreak Perhaps it would be helpful if you used used "Archlinux compatibility" instead of "Archlinux support," but the Cloud & Containers team own this module so it is up to them.

bastelfreak:
Thanks for the response @hunner. I updated the title as you suggested.

@davejrt @scotty-c would it be possible for you to align the way the Modules Team handles those contributions?

ccaum:
Thanks for the contribution @bastelfreak, but @scotty-c is correct. For this module we can only reasonably support a limited number of OSes and we've settled on the ones officially supported by the Docker engine. 

However, I will work with the module's team to determine if we need to have an official policy in place for these kinds of contributions so we can have a more formal process and explanation for how these decisions are made.

Thanks, again.

PR #182, opened 2018-03-12
Docker login Receipt is now removed if login fails and added acceptance tests

### LINKED ISSUE #167: Docker registry is never called after first run
I am encountering issues when docker login is run and a network hiccup occurs (some people love their firewall). The receipt is created either way even though no login has happened. And because of this every pull will fail evens after the network hiccup. 

My proposed solution is to perform the login each time but to surpress the output using unless.

For example:

```
  exec { "${title} auth":
    environment => $auth_environment,
    command     => $auth_cmd,
    user        => $local_user,
    cwd         => '/root',
    path        => ['/bin', '/usr/bin'],
    timeout     => 0,
    unless      => $auth_cmd,
#    refreshonly => $receipt,
  }

```
Pro: we see if the login fails Now in that case it will only show up if something goes wrong.
Con: We dont see the initial (and subsequent logins)

Any ideas on this?

Or add the possibility to just surpress the login in that case both options receipt and continuously login can live together?

sigsegv11:
An attempt to resolve #167 in a clean way

### DIFF
--- a/manifests/registry.pp
+++ b/manifests/registry.pp
@@ -77,6 +77,7 @@ define docker::registry(
 
     # server may be an URI, which can contain /
     $server_strip = regsubst($server, '/', '_', 'G')
+    $_auth_command = "${auth_cmd} || rm -f \"/root/registry-auth-puppet_receipt_${server_strip}_${local_user}\""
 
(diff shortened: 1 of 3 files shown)


davejrt:
LGTM

PR #185, opened 2018-03-13
CLOUD-1637 Readme update

MWilsonPuppet:
This PR makes readme updates as suggested in CLOUD-1637. Types and their parameters have been added. I have not added the classes to the class parameters specified in the readme as this is not consistent with other modules.


davejrt:
LGTM

Issue #183, opened 2018-03-12
Docker image fails to run when specifying cpuset parameter


cjmachado:
Hi,

I'm using docker version 17.12.1-ce and puppetlabs-docker version 1.0.5. Trying to run a docker image with cpuset parameter set to ['0'] fails with the following error message:

```
Notice: /Stage[main]/Docker::Run_instance/Docker::Run[myImage]/Exec[run myImage with docker]/returns: unknown flag: --cpuset
Notice: /Stage[main]/Docker::Run_instance/Docker::Run[myImage]/Exec[run myImage with docker]/returns: See 'docker run --help'.
Error: 'docker run -d --net host -m 1024m --cpuset=0 --detach=true --expose=6060 \
 --expose=6061 \
(...)
```

Running 'docker run --help' shows that there is no --cpuset parameter:

```
docker  run --help

Usage:	docker run [OPTIONS] IMAGE [COMMAND] [ARG...]

Run a command in a new container

Options:
      --add-host list                  Add a custom host-to-IP mapping (host:ip)
  -a, --attach list                    Attach to STDIN, STDOUT or STDERR
      --blkio-weight uint16            Block IO (relative weight), between 10 and 1000, or 0 to disable (default 0)
      --blkio-weight-device list       Block IO weight (relative device weight) (default [])
      --cap-add list                   Add Linux capabilities
      --cap-drop list                  Drop Linux capabilities
      --cgroup-parent string           Optional parent cgroup for the container
      --cidfile string                 Write the container ID to the file
      --cpu-period int                 Limit CPU CFS (Completely Fair Scheduler) period
      --cpu-quota int                  Limit CPU CFS (Completely Fair Scheduler) quota
      --cpu-rt-period int              Limit CPU real-time period in microseconds
      --cpu-rt-runtime int             Limit CPU real-time runtime in microseconds
  -c, --cpu-shares int                 CPU shares (relative weight)
      --cpus decimal                   Number of CPUs
      --cpuset-cpus string             CPUs in which to allow execution (0-3, 0,1)
      --cpuset-mems string             MEMs in which to allow execution (0-3, 0,1)
  -d, --detach                         Run container in background and print container ID
(...)
```
The correct parameter seems to be 'cpuset-cpus'. Thanks!

davejrt:
Fixed in PR #187 and #191

PR #187, opened 2018-03-13
(maint)CLOUD-1768 Fixing incorrect cpuset flag #183

MWilsonPuppet:
This PR is to resolve issue #183. The wrong flag was being passed to set the cpuset. I was not seeing the original warning that was reported but was seeing that when the cpuset parameter was set the container did not run. However when the flag was changed to the correct one the container started as expected. Also updated spec tests.


jeefberkey:
Is this parameter just wrong, or did it change in a recent version?

MWilsonPuppet:
@jeefberkey It looks like the parameter was deprecated in release V1.8.0 and removed in V1.10.0.

jeefberkey:
That's old enough for me :)

davejrt:
LGTM

PR #199, opened 2018-03-16
Limit Image -> Run dependencies to those used

electrofelix:
Use searching in the resource collection to limit adding dependencies on
Docker::Image for only those Docker::Run resources defined to use the
specific image.

This reduces the likelihood of accidental cyclical dependencies where
consumers defined a manifest per service containing an image and run,
and then try to add a dependency from one service class onto another.

### DIFF
--- a/manifests/image.pp
+++ b/manifests/image.pp
@@ -124,4 +124,5 @@ define docker::image(
     }
   }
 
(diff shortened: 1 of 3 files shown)


electrofelix:
Originally reported as garethr/garethr-docker#640 and submitted as garethr/garethr-docker#641

davejrt:
LGTM

PR #200, opened 2018-03-16
Use line breaks and continuation for run flags

electrofelix:
Switch from joining the docker_run_flags using spaces to new lines and
continuations to avoid hitting the glibc MAX_LINE limit used by systemd.

As docker_run_flags is used to provide the args to init scripts and
systemd, switching it to use line continuation and newlines is
sufficient to ensure that systemd/initscritps do not hit any limit
imposed by glibc on a readline and at the same time know to continue
processing onto the next line for the rest of the command to be run.

### DIFF
--- a/lib/puppet/parser/functions/docker_run_flags.rb
+++ b/lib/puppet/parser/functions/docker_run_flags.rb
@@ -82,6 +82,8 @@ module Puppet::Parser::Functions
       flags << param
     end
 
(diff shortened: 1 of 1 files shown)


electrofelix:
Originally submitted as garethr/garethr-docker#695 as a fix for garethr/garethr-docker#660

davejrt:
LGTM

PR #205, opened 2018-03-21
unlocking puppet version for travis

davejrt:


MWilsonPuppet:
@davejrt Looks good to me

Issue #208, opened 2018-03-24
Support Docker Compose file v2.3


XanderXAJ:
## What you expected to happen?
Docker Compose stacks using compose file version 2.3 would be brought up.

I'm using Docker Compose 1.19.0 (provisioned via this plugin), which contains support for compose file v2.3.

## What happened?

I receive the following error:

```
Could not evaluate: Unsupported docker compose file syntax version "2.3"!
```

## How to reproduce it?

Provision any compose v2.3 file, for example:

```
version: '2.3'
services:
  redis:
    image: "redis:alpine"
```

Then try to bring the stack up:

```
docker_compose { '/path/to/the-stack/docker-compose.yml':
  ensure => present,
}
```

## Versions:

```
$ puppet --version
5.4.0
```

```
$ docker version
Client:
 Version:       17.12.1-ce
 API version:   1.35
 Go version:    go1.9.4
 Git commit:    7390fc6
 Built: Tue Feb 27 22:17:53 2018
 OS/Arch:       linux/amd64

Server:
 Engine:
  Version:      17.12.1-ce
  API version:  1.35 (minimum version 1.12)
  Go version:   go1.9.4
  Git commit:   7390fc6
  Built:        Tue Feb 27 22:16:25 2018
  OS/Arch:      linux/amd64
  Experimental: false
```

```
$ facter os
{"name"=>"Ubuntu", "family"=>"Debian", "release"=>{"major"=>"17.10", "full"=>"17.10"}, "lsb"=>{"distcodename"=>"artful", "distid"=>"Ubuntu", "distdescription"=>"Ubuntu 17.10", "distrelease"=>"17.10", "majdistrelease"=>"17.10"}}
```

```
$ puppet module list
/etc/puppetlabs/code/modules-managed
├── ajjahn-samba (v0.5.0)
├── camptocamp-augeas (v1.6.0)
├── camptocamp-postfix (v1.6.0)
├── dhoppe-fail2ban (v1.3.5)
├── ghoneycutt-common (v1.7.0)
├── ghoneycutt-ssh (v3.57.0)
├── herculesteam-augeasproviders_core (v2.1.4)
├── herculesteam-augeasproviders_sysctl (v2.2.0)
├── puppet-extlib (v1.1.0)
├── puppet-letsencrypt (v2.3.0)
├── puppet-unattended_upgrades (v3.1.0)
├── puppetlabs-apt (v4.5.1)
├── puppetlabs-concat (v4.2.0)
[...]

PR #210, opened 2018-03-28
fix for #208

scotty-c:
Change regex to fix the issue raised in #208 This also adds support for Docker compose 3.3 - 6


MWilsonPuppet:
@scotty-c LGTM

Issue #217, opened 2018-04-13
How to use Swarm In Hiera


thatsk:
node.yaml
---
profiles:
- profile::docker

docker::swarm: 
    'cluster_manager':
      init: 'true'
      advertise_addr: '10.174.165.150'
      listen_addr: '10.174.165.150'

My Profile :-
docker.pp

class profile::docker {
  include ::docker
  include ::docker::compose
  include ::docker::run_instance
  include ::docker::registry_auth
}

But when running puppet on node it does not load that. Can anyone let me know what i am missing?

PR #219, opened 2018-04-17
adding in the functionality to define swarm clusters in hiera

davejrt:
Fixes issue #217 

Closes CLOUD-1855

### DIFF
--- a/manifests/swarms.pp
+++ b/manifests/swarms.pp
@@ -1,0 +1,4 @@
+# docker::swarms
+class docker::swarms($swarms) {
+  create_resources(docker::swarm, $swarms)
+}


manifests/swarms.pp
```diff
@@ -0,0 +1,4 @@
+# docker::swarms
+class docker::swarms($swarms) {
+  create_resources(docker::swarm, $swarms)
```
jeefberkey: Create resources in puppet > 4.0?

davejrt: What's your question?

jeefberkey: Iterating over a hash is much more user friendly and safe. In the least, it provides more helpful error messages

PR #221, opened 2018-04-26
Fix incompatiblity with puppet4

nicolasbigler:
By default $version is undef unless a specific version is specified.
In Puppet 4.x regular expression comparisions against non-strings raises
an error.
Thus using the default value for $version will raise an error when using
the registry class in puppet 4.x.
To work around this issue, we first need to check that $version has not
an undef value.


davejrt:
LGTM

PR #227, opened 2018-05-12
Update install.pp to Support Red Hat "yum" Package Provider

jmbelvar81:
Under Red Hat systems using the implemented change in this commit and the attribute *package_source* to *'RedHat'* value I could solve problems to install a docker specific version when you desire use the official repositories. I commented the complete "case" under the *Issue 225*.

Any comment is Welcome about errors described in issue 225 and actual suggested change.


scotty-c:
LGTM

PR #236, opened 2018-05-24
fix "start a container with cpuset" acceptance test on ubuntu1404

mihaibuzgau:
On Ubuntu 14.04 creating containers does not spawn new processes, like it does on Ubuntu 16.04

This PR fixes the assert condition for "the Puppet Docker module clean up before each test docker::run should start a container with cpuset parameter set" acceptance test by changing the assert statement from searching for a process that belongs to a newly spawned container to using the "docker inspect" command.


davejrt:
LGTM

Issue #213, opened 2018-04-06
Adding new users to docker_users results in restart of the docker daemon


pugnacity:
## What you expected to happen?
Add user to docker group without restarting the docker daemon. There is no need for restarting the daemon.
## What happened?
when adding a user to docker_users the docker daemon will be restarted

## How to reproduce it?
add  user to docker_users array which isn't already in the docker group

## Versions:

```
root@selenium02:~# puppet --version
5.5.0
root@selenium02:~# docker version
Client:
 Version:       18.03.0-ce
 API version:   1.37
 Go version:    go1.9.4
 Git commit:    0520e24
 Built: Wed Mar 21 23:10:06 2018
 OS/Arch:       linux/amd64
 Experimental:  false
 Orchestrator:  swarm

Server:
 Engine:
  Version:      18.03.0-ce
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.9.4
  Git commit:   0520e24
  Built:        Wed Mar 21 23:08:35 2018
  OS/Arch:      linux/amd64
  Experimental: false
root@selenium02:~# facter os
{
  architecture => "amd64",
  distro => {
    codename => "stretch",
    description => "Debian GNU/Linux 9.4 (stretch)",
    id => "Debian",
    release => {
      full => "9.4",
      major => "9",
      minor => "4"
    }
  },
  family => "Debian",
  hardware => "x86_64",
  name => "Debian",
  release => {
    full => "9.4",
    major => "9",
    minor => "4"
  },
  selinux => {
    enabled => false
  }
}
```

## Logs:
```
Apr 06 08:43:55 preview01 puppet-agent[27158]: (/Stage[main]/Docker::Config/Docker::System_user[hupli]/Exec[docker-system-user-hupli]/returns) executed successfully
Apr 06 08:43:55 preview01 puppet-agent[27158]: (/Stage[main]/Docker::Service/Exec[docker-systemd-reload-before-service]) Triggered 'refresh' from 1 event
Apr 06 08:43:56 preview01 dockerd[12897]: time="2018-04-06T08:43:56.015208796+02:00" level=info msg="Container 4f67ad205904e439287befaa404506a9124f3e2d4ed15dd20de941f868302ed5 failed to exit within 0 seconds of signal 15 - using the force"
[...]

PR #247, opened 2018-06-18
WIP-Do not merge-CLOUD-1854-Fixing github issue #213

MWilsonPuppet:
This fixes issue #213 where the daemon was being restarted when a new user was added. Spec and acceptance tests have been added also.


spec/defines/system_user_spec.rb
```diff
@@ -11,14 +11,19 @@
...
-
+  
   context 'with default' do
     let(:params) { {'create_user' => true} }
     it { should contain_user('testuser') }
     it { should contain_exec('docker-system-user-testuser').with_command(/docker testuser/) }
     it { should contain_exec('docker-system-user-testuser').with_unless(/grep -qw testuser/) }
   end
 
+  context 'when adding a system user' do
```
davejrt: @MWilsonPuppet we don't need this test. The one you've added in the docker spec is sufficient. This one is failing because you're passing in a param that doesn't actually exist in the define type, not does the exec. A better place to test this would be in the service_spec.rb

spec/classes/docker_spec.rb
```diff
@@ -422,7 +422,6 @@
       it { should compile.with_all_deps }
       it { should contain_class('docker::repos').that_comes_before('Class[docker::install]') }
       it { should contain_class('docker::install').that_comes_before('Class[docker::config]') }
-      it { should contain_class('docker::service').that_subscribes_to('Class[docker::config]') }
```
davejrt: Can we add in as part of this ensuring that the config class comes before the service class?

spec/classes/docker_windows_spec.rb
```diff
@@ -28,7 +28,6 @@
       it { should contain_exec('install-docker-package').with_command("& #{service_install_file}") }
       it { should contain_class('docker::repos').that_comes_before('Class[docker::install]') }
       it { should contain_class('docker::install').that_comes_before('Class[docker::config]') }
-      it { should contain_class('docker::service').that_subscribes_to('Class[docker::config]') }
```
davejrt: Can we add in as part of this ensuring that the config class comes before the service class?

MWilsonPuppet:
@davejrt Cheers for feedback. All changes have been made and unit and acceptance tests passing.

PR #248, opened 2018-06-19
Added architecture to apt source

koshatul:
Fixes the issue with Ubuntu/Debian returning an warning on `apt-get update`

```
N: Skipping acquire of configured file 'stable/binary-i386/Packages' as repository 'https://download.docker.com/linux/ubuntu xenial InRelease' doesn't support architecture 'i386'
```

Because without specifying architecture it attempts to retrieve i386 and amd64.

### DIFF
--- a/manifests/repos.pp
+++ b/manifests/repos.pp
@@ -27,6 +27,7 @@ class docker::repos (
             id     => $package_key,
             source => $key_source,
           },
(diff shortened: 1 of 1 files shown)


koshatul:
My bad, I didn't even see it there, it wasn't in the version I installed, but that's not the latest.

Sorry about that.

PR #250, opened 2018-06-20
Allow custom docker_group and service_name values in run.pp

esalberg:
The default docker_group can be overridden in init.pp, but run.pp only uses the default params values.  Also updated service_name the same way.

I could have moved the parameters into the override specific to docker::run, but I assumed that they were kept more global on purpose.  I'm happy to update the PR if the preference is to override them per docker run.

There are a number of other cases in run.pp where it's pointing at params directly for a parameter that can be overridden in init.  I wanted to keep this use case small / specific to what I needed, but would it make sense to revise those other parameters if an override exists in init.pp?

Please let me know if a spec test is needed.

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -134,8 +134,14 @@ define docker::run(
   }else {
     $docker_command = $docker::params::docker_command
   }
-  $service_name = $docker::params::service_name
-  $docker_group = $docker::params::docker_group
+
+  if defined('docker') {
+    $service_name = $docker::service_name
+    $docker_group = $docker::docker_group
+  } else {
+    $service_name = $docker::params::service_name
+    $docker_group = $docker::params::docker_group
+  }
 
   if $restart {
     assert_type(Pattern[/^(no|always|unless-stopped|on-failure)|^on-failure:[\d]+$/], $restart)
--- a/spec/defines/run_spec.rb
+++ b/spec/defines/run_spec.rb
@@ -3,6 +3,7 @@ require 'spec_helper'
 ['Debian', 'RedHat'].each do |osfamily|
   describe 'docker::run', :type => :define do
     let(:title) { 'sample' }
+    let(:pre_condition) { "class { 'docker': docker_group => 'docker', service_name => 'docker' }" }
     context "on #{osfamily}" do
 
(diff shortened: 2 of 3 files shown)


scotty-c:
@esalberg if you could just make sure the test pass. Then we can help get this merged to master

duritong:
Without this or #272 this module is broken on EL7 with distro packages, as there is no way to set docker_group to 'dockerroot'. which is not automatically chosen by params.pp because `use_upstream_package_source` is by default true.

esalberg:
Sorry about the delay.  I'm assuming that the run spec doesn't know about the values in other manifests, so I moved the docker_group and service_name variables to be parameters and set a default value in the spec.  I'll check the tests when they conclude, but I'll keep working on this in any case.

esalberg:
An update on what I've been working on for the past day:

Rspec has an issue with looking up docker::docker_group because the main Docker class is not included in run.pp.

In init.pp, the main class is ordered before all Docker::Run defines if both are included.
  Class['docker'] -> Docker::Run <||>

However, I didn't want to assume that there weren't cases where the run defines only were configured.  Is it assumed that anyone using the docker::run define will be including the main Docker class?

If so, I believe the code change can be very simple:

include docker
$docker_group = $docker::docker_group
$service_name  = $docker::service_name

In the meantime, I have added an if defined( ) clause in that sets the variables from docker::params if the docker class isn't defined, but that seems a bit hacky.  

Also, in rspec I still had to add a pre_condition including the docker class, which means that the spec tests no longer cover the case where docker::run is called without including docker.

davejrt:
Apologies for the delay in getting back to you. I would be assuming that the main docker class is being included, and that an if defined not be used as I agree it's hacky.

androidkh:
I agree with @davejrt  and propose just a simple bugfix change:
    $service_name = $docker::service_name
    $docker_group = $docker::docker_group

davejrt:
Closing due to inactivity. Feel free to reopn with the suggested changes

esalberg:
I completely forgot about this ticket.  Re-opening  with the requested change.  I also need to add an override of the docker.socket file, but I'll do that in a separate PR.

PR #252, opened 2018-06-25
Run docker compose on windows

florindragos:
- add ability to run docker-compose on windows
- add docker-compose v3 tests that also run on windows


davejrt:
LGTM

Issue #216, opened 2018-04-10
[Feature] - restart container if reported as unhealthy


mlukowicz:
Hello!

Currently docker does not implement restart upon reporting unhealthy (via HEALTHCHECK directive).
It would be nice if we can define container to be restarted if it is misbehaving - meaning running but not doing what we want.

Michał

scotty-c:
@mlukowicz How are you running the container with Puppet?
Would you be able to attach an example

mlukowicz:
I am currently not using swarm.

I simply use docker::run with manage option.

docker::run { 'myapp':
    image               => 'myregistry/myapp',
    ports               => ['5555:5555','6666:6666'],
    expose              => ['5555','6666'],
    restart_service     => true,
    pull_on_start       => true,
  }

mlukowicz:
PS: currently i am doing that by simply defining cron job that extract this info from docker ps output and restart container if necessary - but it would be nice to have that in module.

PR #254, opened 2018-06-25
(CLOUD-1850)-Restart docker container on unhealthy status

MWilsonPuppet:
This PR is to add the functionality requested in #216. This adds the ability to add Docker health check as well as restarting the container if it is reported unhealthy.


davejrt:
LGTM

Issue #255, opened 2018-06-26
Problem installing docker version "17.03.2~ce-0~ubuntu-xenial"


rauschbit:
## What you expected to happen?
We wanted the following docker-ce version to get installed on the defined nodes for our kubernetes environment.
-> `17.03.2~ce-0~ubuntu-xenial`

The recommended Docker version for Kubernetes v1.10 is 17.03.x
-> https://kubernetes.io/docs/imported/release/notes/#external-dependencies

## What happened?
After applying the module to the corresponding node we got an error shown in the log messages below.

## How to reproduce it?
You can reproduce it any time using the docker-ce version as written above.

## Anything else we need to know?
The problem is the regex in the init.pp of the docker module
-> https://github.com/puppetlabs/puppetlabs-docker/blob/master/manifests/init.pp

`if ( $version == undef ) or ( $version !~ /^(17[.]0[0-5][.]\d(~|-|\.)ce|1.\d+)/ ) {`

This regex matches the version we want to install and takes the else to the "docker-engine" installation.
So the wrong apt-repository (apt.dockerproject.org) gets configured + the wrong package (docker-engine) will be tried to get installed with a version which doesn't exist for this package.

## Versions:
```
$ puppet --version
4.10.12

$ docker version
-> we wanted to install this version -> 17.03.2~ce-0~ubuntu-xenial

$ facter os
{
  architecture => "amd64",
  distro => {
    codename => "xenial",
    description => "Ubuntu 16.04.4 LTS",
    id => "Ubuntu",
    release => {
      full => "16.04",
      major => "16.04"
    }
  },
  family => "Debian",
  hardware => "x86_64",
  name => "Ubuntu",
  release => {
    full => "16.04",
    major => "16.04"
  },
  selinux => {
    enabled => false
  }
}

$ puppetlabs-docker version
1.1.0
```

## Logs:
```
Jun 26 09:04:59 test123 puppet-agent[11071]: Could not update: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold --force-yes install docker-engine=17.03.2~ce-0~ubuntu-xenial' returned 100: Reading package lists...
Jun 26 09:04:59 test123 puppet-agent[11071]: Building dependency tree...
Jun 26 09:04:59 test123 puppet-agent[11071]: Reading state information...
[...]

PR #257, opened 2018-06-28
fixes regex for versions in 2017 with a release not 0 or 1

davejrt:
fixes issue #255


MWilsonPuppet:
@davejrt looks good to me.

Issue #243, opened 2018-05-30
docker::stack documents resolve_image without actually implementing it


joernott:
## What you expected to happen?

```
      docker::stack{ $stack_name:
        ensure        => present,
        resolve_image => always,
        stack_name    => $stack_name,
        compose_file  => $compose_file,
        require       => [Class['docker'],File[$compose_file]],
      }
```
should work.

## What happened?
Despite this parameter being documented in the module, it is not implemented. I am getting the following error:
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Docker::Stack[test]: has no parameter named 'resolve_image'

## How to reproduce it?
Use the documented "resolve_image" parameter 

## Versions:
1.1.0 up to latest at the time of writing (https://github.com/puppetlabs/puppetlabs-docker/commit/b21706029ae976e92960b56153f83d41d2ac2b64)
```
$ puppet --version
4.10.11

$ docker version
Docker version 18.05.0-ce, build f150324

$ facter os
{
  architecture => "x86_64",
  family => "RedHat",
  hardware => "x86_64",
  name => "CentOS",
  release => {
    full => "7.5.1804",
    major => "7",
    minor => "5"
  },
  selinux => {
    enabled => false
  }
}


$ puppet module list
/etc/puppetlabs/code/environments/production/modules (no modules installed)
/etc/puppetlabs/code/modules (no modules installed)
/opt/puppetlabs/puppet/modules (no modules installed)

```

## Logs:
```
Info: Using configured environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Docker::Stack[test]: has no parameter named 'resolve_image' at /etc/puppetlabs/code/production/modules/oc/manifests/docker.pp:97 on node docker1.xxxxxx
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```

davejrt:
A ticket has been raised internally. An update will be made here when it's picked up in our sprint

PR #261, opened 2018-07-05
CLOUD-1967-Fixing -Docker_stacks-image resolve

MWilsonPuppet:
This fixes issue #243 where image_resolve was documented but not implemented. I have implemented this and updated spec test. I also verified it was working as expected and that the stack was created successfully.


davejrt:
LGTM

PR #262, opened 2018-07-06
Configure Docker Swarm on Windows

florindragos:
- add ability to create a docker swarm on windows (firewall ports need to be opened)
- add some docker swarm acceptance tests in a different test suite since it takes a while to run on windows and needs 2 vms


davejrt:
LGTM

PR #264, opened 2018-07-10
fixes fact for windows

davejrt:
updates the docker command to use powershell on windows due to windows cmd not accepting quotes


mihaibuzgau:
+1

PR #272, opened 2018-07-12
Read the docker_group parameter from the docker class.

hlaf:


duritong:
Without this or #250 this module is broken on EL7 with distro packages, as there is no way to set docker_group to 'dockerroot'. which is not automatically chosen by params.pp because `use_upstream_package_source` is by default true.

davejrt:
This module only supports official docker releases, not distro packages.

Issue #211, opened 2018-03-29
systemd service runs always the latest/master image


koma85:
## What you expected to happen?
when I specify a tag for an Image the systemd service (not checked the other) should append the :tag to the script. This does not happen.
## What happened?
The systemd running script is wrong

## How to reproduce it?
place a tag and look for what images is used.

## Anything else we need to know?
Probably just need a patch on the ERB

## Versions:

```
This is not version related.
```

## Logs:
```
This is not visible from logs.
```

scotty-c:
@koma85 can you send through a copy of the systemd file that this is happening on

koma85:
Calling an image that is called centreon_vmware_connector with tag fix-division-by-0

This is the file 

> /etc/systemd/system/multi-user.target.wants/docker-centreon_vmware_connector.service

```
# This file is managed by Puppet and local changes
# may be overwritten

[Unit]
Description=Daemon for centreon_vmware_connector
After=docker.service
Wants=
Requires=docker.service

[Service]
Restart=on-failure
StartLimitInterval=20
StartLimitBurst=5
TimeoutStartSec=0
RestartSec=5
Environment="HOME=/root"
SyslogIdentifier=docker-centreon_vmware_connector
ExecStartPre=-/usr/bin/docker kill centreon_vmware_connector
ExecStartPre=-/usr/bin/docker rm  centreon_vmware_connector

ExecStart=/usr/bin/docker run \
        --net centreon -m 0b --restart=always \
        --name centreon_vmware_connector \
        git.XXXYYZZ.com:8443/containers/centreon-vmware-connector \

ExecStop=-/usr/bin/docker stop --time=0 centreon_vmware_connector
ExecStop=-/usr/bin/docker rm  centreon_vmware_connector

[Install]
WantedBy=multi-user.target
```

should be:
```
# This file is managed by Puppet and local changes
# may be overwritten

[Unit]
Description=Daemon for centreon_vmware_connector
After=docker.service
Wants=
Requires=docker.service

[Service]
Restart=on-failure
StartLimitInterval=20
StartLimitBurst=5
TimeoutStartSec=0
RestartSec=5
Environment="HOME=/root"
SyslogIdentifier=docker-centreon_vmware_connector
ExecStartPre=-/usr/bin/docker kill centreon_vmware_connector
ExecStartPre=-/usr/bin/docker rm  centreon_vmware_connector

ExecStart=/usr/bin/docker run \
[...]

PR #277, opened 2018-07-17
Update README.md

davejrt:
adding in a note to advise against using latest, or not specifying a tag at all is against best practise as raised in #211


mihaibuzgau:
👍

PR #279, opened 2018-07-17
CLOUD-2018-Add-health-check-interval

MWilsonPuppet:
This PR adds the ability to set the Docker health check interval time. I verified that the interval time was being set as expected. It also fixes a typo in the readme for the health check command.


lib/puppet/parser/functions/docker_run_flags.rb
```diff
@@ -50,6 +50,10 @@ module Puppet::Parser::Functions
       flags << "--health-cmd #{opts['health_check_cmd']}"
     end
 
+    if opts['health_check_interval'].to_i != 30
```
davejrt: we need this set to undef and not to set a default for the healthcheck cmd. At the moment this appears to evaluate as true if someone uses a value other than 30s. But, if someone does want their check at 30s then they won't actually be able to do it.

MWilsonPuppet: @davejrt yes makes complete sense Dave Ill make changes now. Cheers

PR #283, opened 2018-07-18
fixes restarting containers with changes to run arguments

davejrt:
this PR fixes an issue where the container would not be restarted with new arguments from the docker run type, and the exec was not triggered by the new run script.


MWilsonPuppet:
Looks good to me Dave.

PR #284, opened 2018-07-18
Health fix

davejrt:
fixes issue where health check commands with spaces were not captured in quotes, and stop contaienrs from starting due to an error in the docker run command


MWilsonPuppet:
Looks good to me Dave

PR #288, opened 2018-07-20
[CLOUD-2033] determine if a compose service extends another compose service

mihaibuzgau:
Take into consideration the image of an extended service when determining if a service exists.


davejrt:
LGTM

Issue #241, opened 2018-05-30
Puppet downloads new Docker image when running in noop mode


jethrocarr:
**Fault**
When running Puppet with `--noop` it seems that this module will still pull Docker images

**Expectation**
`--noop` should be a no-impact mode and should not result in Docker images being pulled.

**Reproducing**
To reproduce, have a manifest that defines a `docker::image`, eg:
```
docker::image { 'ubuntu':
   image_tag => 'latest',
}
```

Then run puppet with `---noop`

davejrt:
I can't replicate your issue. With the following manifest initially I get docker images as you'd expect
```
node default {
  include docker
}
```

```
root@node-01:~# docker images
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
root@node-01:~#
```

If I update the manifest to the following then run with noop I get the following error
```
node default {
  include docker

 docker::image { 'alpine':
   image_tag => 'latest',
}
```

```
root@node-01:/tmp# cat manifests-a11d1078b1b1f2e3bdea27312f6ba513/default.pp
node default {
  include docker

  docker::image {'alpine':
   image_tag => 'latest',
  }
}

root@node-01:/tmp# puppet apply --noop manifests-a11d1078b1b1f2e3bdea27312f6ba513/default.pp --modulepath /tmp/modules
Warning: ModuleLoader: module 'docker' has unresolved dependencies - it will only see those that are resolved. Use 'puppet module list --tree' to see information about modules\n   (file & line not available)
Notice: Compiled catalog for node-01.home in environment production in 0.34 seconds
Notice: /Stage[main]/Main/Node[default]/Docker::Image[alpine]/File[/usr/local/bin/update_docker_image.sh]/ensure: current_value 'absent', should be 'present' (noop)
Error: /Stage[main]/Main/Node[default]/Docker::Image[alpine]/Exec[echo 'Update of alpine:latest complete']: Could not evaluate: Could not find command '/usr/local/bin/update_docker_image.sh'
Notice: Applied catalog in 0.27 seconds
root@node-01:/tmp# docker images
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
root@node-01:/tmp# puppet apply manifests-a11d1078b1b1f2e3bdea27312f6ba513/default.pp --modulepath /tmp/modules
[...]

PR #297, opened 2018-07-25
[CLOUD-1981] docker image update script runs in noop mode

mihaibuzgau:
The --noop flag tells Puppet to determine which resources are out of sync, and to report them without actually synchronizing them.  The 'onlyif' and 'unless' commands of an Exec are used in the process of determining whether the Exec is already in sync, therefore they must be run during a --noop Puppet run. 

This PR changes the 'onlyif' command for ensure == 'latest' or $image_tab == 'latest' from a script that takes action (downloads images) to a command that only check the state. 

Fixes: https://github.com/puppetlabs/puppetlabs-docker/issues/241

### DIFF
--- a/manifests/image.pp
+++ b/manifests/image.pp
@@ -140,17 +140,7 @@ define docker::image(
       timeout     => $exec_timeout,
       logoutput   => true,
     }
-  } elsif $ensure == 'latest' or $image_tag == 'latest' {
-    exec { "echo 'Update of ${image_arg} complete'":
-      environment => $exec_environment,
-      path        => $exec_path,
-      timeout     => $exec_timeout,
-      onlyif      => $image_install,
-      require     => File[$update_docker_image_path],
-      provider    => $exec_provider,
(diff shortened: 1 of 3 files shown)


davejrt:
LGTM

coder-hugo:
This change makes the value `latest` for the `ensure` of a `docker::image` useless since the `$image_install` is only executed if the image in not already installed (`unless      => $_image_find`). So you will never get a never version of the specified installed unless you change the `image_tag`

davejrt:
@coder-hugo we are aware this was a breaking change. This was reflected in the documentation as well as in the version numbers, and we discussed are reasons in issue #316.

coder-hugo:
IMHO such a breaking change should be noted within the changelog and not somewhere hidden in the README. I'd expect that most user will just have a look to the changelog before updating a module and they will not read through the whole README to find some breaking changes.

davejrt:
We'll take your feedback on board for any future breaking changes.

PR #299, opened 2018-07-30
Fix for registry password not being inserted due to single quotes

ConorPKeegan:
The insertion of the single quotes before the 2.0.0 release has caused anything that uses the registry to no longer work. With the single quotes you end up with the login command looking like:
```
docker login -u '<USERNAME>' -p '${password}' <SERVER>
```

The problem is with `'$password'`, as single quotes prevent the variable from being changed into the environment variable. You can only insert like this with double quotes.


florindragos:
👍

bramblek1:
Thanks for finding that @ConorPKeegan - this drove me totally crazy.

Issue #289, opened 2018-07-20
Cannot install Docker 17.03.2 on CentOS


azbpa:
## What you expected to happen?
I want to install Docker 17.03.**2** on CentOS.

## What happened?
```
Error: Could not update: Failed to update to version 17.03.2.ce, got version 17.03.1.ce-1.el7.centos instead
Error: /Stage[main]/Docker::Install/Package[docker]/ensure: change from 17.03.1.ce-1.el7.centos to 17.03.2.ce failed: Could not update: Failed to update to version 17.03.2.ce, got version 17.03.1.ce-1.el7.centos instead
```

## How to reproduce it?
```
class { 'docker':
  version => '17.03.2.ce',
}
```

## Anything else we need to know?
To me it seems that the following statements switches between different RPM repositories:
https://github.com/puppetlabs/puppetlabs-docker/blob/1.1.0/manifests/init.pp#L543

So for Docker version 17.03.**2** it uses: https://yum.dockerproject.org/repo/main/centos/7/Packages/
However, the desired version is not available there, but it would be available on the default repository here: https://download.docker.com/linux/centos/7/x86_64/stable/Packages/

So why the module is using the old repository for a version prior to 17.06 and not prior to 17.03?

## Versions:

Puppet version: 4.10.11
Module version: 1.1.0
CentOS version: 7.4

davejrt:
Someone will look into this. If you look at the different repos between the OS we support, they are not evenly split between versions and package names, which is why we drew a line in the sand at 17.05. There appear to be edge cases that we didn't account for at the time, so we may need to change our logic. 

Thanks for bringing this to our attention.

mihaibuzgau:
Hi @azbpa 
This PR: https://github.com/puppetlabs/puppetlabs-docker/pull/300 allows you to deploy Docker version 17.03.2.ce on CentOS 7.
Please note that this version of docker has been deprecated and to install it you must provide the `--setopt=obsoletes=0` parameter. Example:
```
class { 'docker':
  version => '17.03.2.ce',
  repo_opt => '--setopt=obsoletes=0',
}
```

PR #300, opened 2018-07-31
[CLOUD-2040]enable extras repo only in CentOS 6 or lower

mihaibuzgau:
This PR disables "extras" CentOS repo for versions greater and equal to 7.
In CentOS 7 and above the packages from the extras repo where tested by the CentOS developers and the repo has been enabled by default: https://wiki.centos.org/AdditionalResources/Repositories 

Passing the --enablerepo=extras throws a repository not found error: `Error getting repository data for extras, repository not found`. 

This PR fixes the following issue: https://github.com/puppetlabs/puppetlabs-docker/issues/289

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -215,8 +215,6 @@ class docker::params {
       # repo_opt to specify install_options for docker package
       if $::operatingsystem == 'RedHat' {
         $repo_opt = '--enablerepo=rhel-7-server-extras-rpms'
-      } elsif $::operatingsystem == 'CentOS' {
-        $repo_opt = '--enablerepo=extras'
       } else {
         $repo_opt = undef
       }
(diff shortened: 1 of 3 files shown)


davejrt:
@mihaibuzgau given we don't support anything earlier than Centos7 I think it makes more sense just to get rid of it altogether.

manifests/params.pp
```diff
@@ -215,7 +215,7 @@
       # repo_opt to specify install_options for docker package
       if $::operatingsystem == 'RedHat' {
         $repo_opt = '--enablerepo=rhel-7-server-extras-rpms'
-      } elsif $::operatingsystem == 'CentOS' {
+      } elsif ($::operatingsystem == 'CentOS') and (Integer($facts["os"]["release"]["major"]) < 7) {
```
davejrt: I think we can just get rid of this altogether

davejrt:
LGTM

PR #301, opened 2018-08-01
removing keyring packages, no longer required 

davejrt:
and causing a dependency cycle


MWilsonPuppet:
Looks good to me Dave.

PR #302, opened 2018-08-01
adds in exec unit test for addition of env vars into exec define

davejrt:


MWilsonPuppet:
looks good to me Dave

PR #305, opened 2018-08-02
allow to trigger deploy with notify

spali:
this allows to trigger a deploy the resource i.e. by a file resource notify.

### DIFF
--- a/manifests/stack.pp
+++ b/manifests/stack.pp
@@ -50,31 +50,36 @@ define docker::stack(
   $docker_command = "${docker::params::docker_command} stack"
 
   if $ensure == 'present'{
-      $docker_stack_flags = docker_stack_flags ({
+    $docker_stack_flags = docker_stack_flags ({
       stack_name => $stack_name,
       bundle_file => $bundle_file,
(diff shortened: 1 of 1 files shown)


spali:
HI @davejrt 
The change does only have an effect if you trigger a notify on the `docker::stack` resource.
I.e.
```
file { '/mpt/my_stack.yaml':
  content => '...'
  notify => Docker::Stack['my_stack']
}
docker::stack { 'my_stack':
...
}
```
If the file changes, the stack get updated. I made it this way to not change as less as possible, because.
A cleaner solution (but with breaking changes) would be to have the `file` resource in the `docker::stack` definition.

davejrt:
@spali understood, thanks. Would mind updating the documentation to reflect your change and ensuring unit tests are passing?

spali:
The current solution is somehow shaky, some failed deploys just return successfully :( 
Currently not a lot of time to fixing this, also because I switched to kubernetes and using the puppetlabs-kubernetes module.
So I close this pull request. If someone has time, feel free to reuse the work.

PR #306, opened 2018-08-05
CLOUD-2046-Add logic for root dir flag

MWilsonPuppet:
This PR adds logic to use the new --data-root flag for versions > 17.05 and the -g flag for versions <= 17.05


davejrt:
LGTM

PR #308, opened 2018-08-08
Do not manage /etc/sysconfig/docker-storage on RedHat based systems.

duritong:
On RedHat based systems docker-storage should be managed through
/etc/sysconfig/docker-storage-setup, which will then configure
/etc/sysconfig/docker-storage on the daemon startup. However,
by managing both on RedHat based systems, we constantly wrote it
based on the template content, restarted the docker daemon,
which then overwrote the file back so on the next puppet run the
same thing happened again.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -173,7 +173,9 @@ class docker::params {
     }
     'RedHat' : {
       $service_config = '/etc/sysconfig/docker'
-      $storage_config = '/etc/sysconfig/docker-storage'
+      # redhat based systems shall configure their storage over
+      # /etc/sysconfig/docker-storage-setup
+      $storage_config = undef
       $storage_setup_file = '/etc/sysconfig/docker-storage-setup'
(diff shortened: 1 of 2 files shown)


duritong:
As I wrote, using the docker daemon shipped by RedHat creates the docker-storage file, through the docker-storage service which is a dependency for the docker service. This service uses the docker-storage-setup file as an input and the docker-storage file as output. Read here for background: https://github.com/projectatomic/container-storage-setup/blob/master/container-storage-setup.1

This might be different for what Docker inc. ships, but not if you don't go with a third party module.

scotty-c:
@duritong This module only supports Docker from Docker inc. It does not support Project Atomic.
I am sorry we will not be merging this PR

duritong:
Well you allow to disable usage of third party repos

duritong:
and you are still managing /etc/sysconfig/docker-storage-setup which is kind of double then

PR #324, opened 2018-09-06
add custom_unless for docker::run exec

esalberg:
Rancher uses an initial Docker run command that exits after spawning other containers.  This allows configuring the run exec not to execute if the other named container is running.

(It looks like it's not necessary to change the onlyif on the start exec, but I'll put in a second PR if needed.)

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -72,6 +72,10 @@
 # configuration.  Most commonly used are on-failure or always.
 # Default: on-failure
 #
+# [*custom_unless*]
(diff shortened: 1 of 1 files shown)


esalberg:
I couldn't find a clear answer online as to whether [] resolves to Boolean false or true (and didn't have time to test), so I went with the most everything resolves to true and set the default to an explicit undef.  I can change it to the empty array if preferred.

davejrt:
I am happy enough with undef as the default value in this case.

PR #325, opened 2018-09-07
Fix error messages from docker facts if docker not running

stdietrich:
If the docker packages are installed on a machine, but the docker daemon is not running, the docker and docker_server_version fact print error messages:

```
# puppet agent -t
.....
Error: Facter: error while resolving custom fact "docker": A JSON text must at least contain two octets!
Error: Facter: error while resolving custom fact "docker_server_version": undefined method `[]' for nil:NilClass
```

The server values are not present in the json output.

This PR changes this, so that the JSON error messages are suppressed.

### DIFF
--- a/lib/facter/docker.rb
+++ b/lib/facter/docker.rb
@@ -21,7 +21,11 @@ end
 Facter.add(:docker_server_version) do
   setcode do
     docker_version = Facter.value(:docker_version)
-    docker_version['Server']['Version'] if docker_version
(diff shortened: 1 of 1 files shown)


davejrt:
Can you ensure your unit tests are passing?

stdietrich:
I have updated my PR with the small change and unit tests are now passing.

davejrt:
LGTM

PR #326, opened 2018-09-10
Fix windows default paths

florindragos:
read windows system paths from custom facts


davejrt:
@glennsarti would be quite keen to get your input here

manifests/image.pp
```diff
@@ -38,11 +38,11 @@
 
   if $::osfamily == 'windows' {
     $update_docker_image_template = 'docker/windows/update_docker_image.ps1.erb'
-    $update_docker_image_path = 'C:/Windows/Temp/update_docker_image.ps1'
-    $exec_environment = 'PATH=C:/Program Files/Docker/'
+    $update_docker_image_path = "${::systemroot}/Temp/update_docker_image.ps1"
```
glennsarti: See above. Systemroot\temp is not what you want.

manifests/params.pp
```diff
@@ -25,11 +25,11 @@
   $tls_enable                        = false
   $tls_verify                        = true
   if ($::osfamily == 'windows') {
-    $tls_cacert                        = 'C:/ProgramData/docker/certs.d/ca.pem'
-    $tls_cert                          = 'C:/ProgramData/docker/certs.d/server-cert.pem'
-    $tls_key                           = 'C:/ProgramData/docker/certs.d/server-key.pem'
+    $tls_cacert                        = "${::program_data_path}/docker/certs.d/ca.pem"
```
glennsarti: This is configurable for the Docker daemon.  Should this path be parameterised?

manifests/registry.pp
```diff
@@ -48,9 +48,9 @@
   $docker_command = $docker::params::docker_command
 
   if $::osfamily == 'windows' {
-    $exec_environment = ['PATH=C:/Program Files/Docker/']
+    $exec_environment = ["PATH=${::program_files_path}/Docker/"]
```
glennsarti: Setting PATH and `exec_path` is bad practice

https://puppet.com/docs/puppet/5.3/types/exec.html#exec-attribute-environment

lib/facter/docker.rb
```diff
@@ -67,3 +67,24 @@ def interfaces
     end
   end
 end
+
+Facter.add(:systemroot) do
```
glennsarti: I'm not keen on these fact names.  It's very possible they will collide with other modules that need to do same thing.

Perhaps prefix with `docker_` or something?

Also it's generally considered better practice for one file per fact. Although I see prior art says otherwise.

manifests/run.pp
```diff
@@ -218,11 +218,11 @@
...
 
   if $::osfamily == 'windows' {
-    $exec_environment = 'PATH=C:/Program Files/Docker/;C:/Windows/System32/'
+    $exec_environment = "PATH=${::program_files_path}/Docker/;${::systemroot}/System32/"
     $exec_timeout = 3000
-    $exec_path = ['c:/Windows/Temp/', 'C:/Program Files/Docker/']
+    $exec_path = ["${::systemroot}/Temp/", "${::program_files_path}/Docker/"]
     $exec_provider = 'powershell'
-    $cidfile = "c:/Windows/Temp/${service_prefix}${sanitised_title}.cid"
+    $cidfile = "${::systemroot}/Temp/${service_prefix}${sanitised_title}.cid"
```
glennsarti: See above. Systemroot\temp is not what you want.

manifests/services.pp
```diff
@@ -106,9 +106,9 @@
   }
 
   if $::osfamily == 'windows' {
-    $exec_environment = 'PATH=C:/Program Files/Docker/'
+    $exec_environment = "PATH=${::program_files_path}/Docker/"
     $exec_timeout = 3000
-    $exec_path = ['c:/Windows/Temp/', 'C:/Program Files/Docker/']
+    $exec_path = ["${::systemroot}/Temp/", "${::program_files_path}/Docker/"]
```
glennsarti: See above about PATH= and `exec_path`

templates/windows/update_docker_image.ps1.erb
```diff
@@ -4,22 +4,25 @@
 # Returns 3 if something when wrong.
 #
 
-$docker_image=$args[0]
-$docker_cmd = "C:\Program Files\Docker\docker.exe"
+param
+(
+    [string]$DockerImage
+)
+$docker_cmd = "${Env:ProgramFiles}\docker\docker.exe"
```
glennsarti: Given docker_cmd is part of the params.pp surely we should be using that?

spec/defines/image_windows_spec.rb
```diff
@@ -9,42 +9,45 @@
...
     context 'with ensure => present and docker_file => Dockerfile' do
         let(:params) { { 'ensure' => 'present', 'docker_file' => 'Dockerfile' } }
-        it { should contain_exec('Get-Content Dockerfile | docker build -t base -') }
+        it { should contain_exec('Get-Content Dockerfile -Raw | docker build -t base -') }
     end
 
     context 'with ensure => present and image_tag => nanoserver' do
         let(:params) { { 'ensure' => 'present', 'image_tag' => 'nanoserver' } }
-        it { should contain_exec('& C:/Windows/Temp/update_docker_image.ps1 base:nanoserver') }
+        it { should contain_exec('& C:/Windows/Temp/update_docker_image.ps1 -DockerImage base:nanoserver') }
```
glennsarti: These tests seem very brittle for the resource name.  Perhaps a better resource name would make sense?

glennsarti:
@florindragos So you're running your Windows tests under cygwin?

glennsarti:
Seems very odd given a lot of other modules use env vars in acceptance.  Have you got some debug logs etc. to show the issue?

florindragos:
I'm seeing really strange behaviour using TMP env var for the custom fact:
- isting the fact on the VM (beaker does cmd.exe /c):

```
bash-4.3$ cmd.exe /c puppet facts | grep docker_user_temp_path
    "docker_user_temp_path": "C:\\cygwin64\\tmp",
```
but running the tests:
Warning: Unknown variable: '::docker_user_temp_path'.

even if we got C:\\cygwin64\\tmp, this is not the value we are expecting...

glennsarti:
On a real windows machine using PowerShell

Note - Ignore the `Get http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.38/version:` errors. My docker VM wasn't running at the time.

Using `"#{ENV['LOCALAPPDATA']}\\Temp"`
``` text
C:\Source\tmp\puppetlabs-docker-pr326 [master ↑2 +0 ~1 -0 !]> & bundle exec puppet facts --modulepath .\spec\fixtures\modules\ | sls docker_user_temp_path
Failed to load feature test for cfpropertylist: cannot load such file -- CFPropertyList
error during connect: Get http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.38/version: open //./pipe/docker_engine: The system cannot find the file specified. In the default daemon configuration on Windows, the docker client must be run elevated to connect. This error may also indicate that the docker daemon is not running.
error during connect: Get http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.38/info: open //./pipe/docker_engine: The system cannot find the file specified. In the default daemon configuration on Windows, the docker client must be run elevated to connect. This error may also indicate that the docker daemon is not running.
Warning: Facter: Fact resolution fact='puppet_classes', resolution='<anonymous>' resolved to an invalid value: Expected all to be one of [Integer, Float, TrueClass, FalseClass, NilClass, String, Array, Hash], but was Symbol

    "docker_user_temp_path": "C:\\Users\\glenn.sarti\\AppData\\Local\\Temp",
```

Using `ENV['TMP']
``` text
C:\Source\tmp\puppetlabs-docker-pr326 [master ↑2 +0 ~1 -0 !]> & bundle exec puppet facts --modulepath .\spec\fixtures\modules\ | sls docker_user_temp_path
Failed to load feature test for cfpropertylist: cannot load such file -- CFPropertyList
error during connect: Get http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.38/version: open //./pipe/docker_engine: The system cannot find the file specified. In the default daemon configuration on Windows, the docker client must be run elevated to connect. This error may also indicate that the docker daemon is not running.
error during connect: Get http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.38/info: open //./pipe/docker_engine: The system cannot find the file specified. In the default daemon configuration on Windows, the docker client must be run elevated to connect. This error may also indicate that the docker daemon is not running.
Warning: Facter: Fact resolution fact='puppet_classes', resolution='<anonymous>' resolved to an invalid value: Expected all to be one of [Integer, Float, TrueClass, FalseClass, NilClass, String, Array, Hash], but was Symbol

    "docker_user_temp_path": "C:\\Users\\GLENN~1.SAR\\AppData\\Local\\Temp",
```

Note that these give different values

So given that;
* You get different behaviour between manual and beaker testing
* 99.999999% of windows users do not use cygwin SSH shells
* Works fine on real Windows
* There is a difference between TMP and LOCALAPPDAT (8.3 filenames), although this is minor

Changing your code because of a flaky test or test environment is bad idea.  We know that the correct env var is TMP, and the fact you're working around this is a "bad smell" in the testing framework.

florindragos:
@glennsarti you're right. making assumptions for the temp folder location in the module just so the tests are working is wrong. The temp folder could also be C:\\Users\\Administrator\\AppData\\Local\\Temp\\2
I changed the custom fact to ENV['TEMP'] and explicitly setting the TEMP variable in the spec_helper_acceptance.ro when running the acceptance tests seems to work just fine.

spec/spec_helper_acceptance.rb
```diff
@@ -130,11 +130,10 @@ def retry_on_error_matching(max_retry_count = 3, retry_wait_interval_secs = 5, e
         end
 
         if fact_on(host, 'osfamily') == 'windows'
-          win_host = only_host_with_role(hosts, 'default')
-          @windows_ip = win_host.ip
-          apply_manifest_on(host, "class { 'docker': docker_ee => true, extra_parameters => '\"insecure-registries\": [ \"#{@windows_ip}:5000\" ]' }")
+          apply_manifest_on(host, "class { 'docker': docker_ee => true }")
           docker_path = "/cygdrive/c/Program Files/Docker"
           host.add_env_var('PATH', docker_path)
+          host.add_env_var('TEMP', 'C:\Users\Administrator\AppData\Local\Temp')
```
glennsarti: This still seems very wrong.  The host should already have a correctly setup TEMP directory.  

Even though the cygwin environment says `\tmp`, external processes created from within cygwin use `C:\cygwin64\tmp`

Windows 2012R2 Pooler VM

``` text
Administrator@adx7v3xdcx8nwfn ~
$ export
...
declare -x TEMP="/tmp"
declare -x TERM="xterm-256color"
declare -x TMP="/tmp"
...

Administrator@adx7v3xdcx8nwfn ~
$ cmd.exe /c set temp
TEMP=C:\cygwin64\tmp

Administrator@adx7v3xdcx8nwfn ~
$ powershell.exe -Command "Write-Host \$ENV:TEMP"
C:\cygwin64\tmp
```

florindragos: That is the way it should work. Even though facter shows the fact pointing to `C:\cygwin64\tmp`,
when running tests, the fact does not seem to be initialized:
`Warning: Unknown variable: '::docker_user_temp_path'`
This happens if I don't set the TEMP env var from within the spec_helper_acceptance...
This works fine for the other 2 facts that are being read from the environment.

RandomNoun7: It's difficult to understand how overriding a system set env variable is the right way to go, especially since it's being hard coded to a specific user's temp path that may or may not be the correct user for a particular run.

Once this variable is overwritten in this way, it's hard to tell what might be broken down the road. Doesn't it seem like this issue you are seeing might be down to misbehavior on the part of beaker or cygwin, that should be worked around more formally?

lib/facter/docker.rb
```diff
@@ -1,6 +1,34 @@
...
+  confine :osfamily => :windows
+  setcode do
+    ENV['ProgramData']
+  end
+end
+
+Facter.add(:docker_user_temp_path) do
+  confine :osfamily => :windows
+  setcode do
+    ENV['TEMP']
```
Iristyle: For the Puppet service this is probably OK. For end users with Unicode characters in their usernames running `puppet apply`, there's a chance that this corrupts the directory name due to Ruby bugs.

If you must expose the TEMP directory as a fact (which I would generally advise against if you can help it), you'll want to use the built-in Puppet helpers `Puppet::Util.get_env` - see https://github.com/puppetlabs/puppet/blob/24ead48f617cd3912491fe419ac7b67cda53a320/lib/puppet/util.rb#L40-L52

spec/spec_helper_acceptance.rb
florindragos: Unfortunately, I have not found a way to work around this.
After doing some more digging around, it looks like the user profile is not being loaded when beaker runs the commands.
The commands are being run remote through ssh: `ssh Administrator@host -i key cmd /c puppet apply /tmp/manifest`

Logging in with ssh will load your user profile and give you the TMP env var but running it remotely will not:

```
ssh Administrator@e34ei8s1mokqrsl.delivery.puppetlabs.net -i .ssh/id_rsa-acceptance
-bash-4.3$ cmd /c set TMP
TMP=C:\cygwin64\tmp
```
vs
```
ssh Administrator@e34ei8s1mokqrsl.delivery.puppetlabs.net -i .ssh/id_rsa-acceptance cmd /c set TMP
Environment variable TMP not defined
```

Adding the variable in `.ssh/environment` would probably fix this, but we still need to enter the value manually.

I agree, it's not right to hardcode the username but since we are running on the vmpooler VMs, we can assume it is running under Administrator.

lib/facter/docker.rb
florindragos: Thanks for the tip. I'll update the code with `Puppet::Util.get_env`.
Is there a reason why we should not use expose the TEMP directory as a custom fact and is there a better alternative to getting to the temporary directory from within a manifest?

PR #327, opened 2018-09-10
More idiomatic usage of parens for 'fail translate(('

PierreR:
We are using [language-puppet](https://github.com/bartavelle/language-puppet) and need to maintain a forge just because of this.


davejrt:
LGTM

Issue #108, opened 2018-01-09
Run fails if you have existing networks with no IPAM settings


krjensen:
Hi,

If you use puppet to create networks, the run fails if you have an existing network created without any IPAM driver specified. 

To fix this you have to change the compose from:
```
networks:
  esnet:
```
to
```
networks:
  esnet:
    ipam:
      driver: default
```

davejrt:
hi @krjensen can you please provide the following:

Puppet version
OS name/version
Code sample to reproduce the issue?

Thanks in advance

davejrt:
Closing due to inactivity

nurriz:
Puppet version: 4.10.11
OS: Ubuntu 16.04.3 LTS

Bring up a stack with this compose:
```
version: '3.4'
services:
  web:
    image: nginx
    networks:
      test:
networks:
  test:
```
docker stack deploy -c docker-compose.yml test
This creates a test_test network that makes the docker_network type fail.
`puppet resource docker_network`
yields

> Error: Could not run: undefined method `empty?' for nil:NilClass

mihkelader:
OS: CentOS 7.5
Puppet: 5.5.3
Docker: 18.03.1-ce

Caused by:
`lib/puppet/provider/docker_network/ruby.rb:41
subnet = unless obj['IPAM']['Config'].empty?`

Code assumes that IPAM -> Config is never null

davejrt:
Thanks for the extra info. Someone will look into it

davejrt:
I am yet to be able to recreate this problem. Can you show me a docker inspect of the network that's causing the issue?

krjensen:
The important thing to notice here, is that the network causing the problem is created WITHOUT puppet, with the following command: `docker network create test --driver=overlay`

```
docker inspect test
[
    {
        "Name": "test",
        "Id": "3pgdij03orfhfw35gy8msm94m",
        "Created": "0001-01-01T00:00:00Z",
        "Scope": "swarm",
        "Driver": "overlay",
        "EnableIPv6": false,
        "IPAM": {
            "Driver": "default",
            "Options": null,
            "Config": []
        },
        "Internal": false,
        "Attachable": false,
        "Ingress": false,
        "ConfigFrom": {
            "Network": ""
        },
        "ConfigOnly": false,
        "Containers": null,
        "Options": {
            "com.docker.network.driver.overlay.vxlanid_list": "4221"
        },
        "Labels": null
    }
]
```

[...]

PR #329, opened 2018-09-13
fixes puppet run failures with no IPAM driver

davejrt:
fixes issue #108


MWilsonPuppet:
LGTM

tbrouhier:
> fixes issue #108

Hi, 

I don't think it does as the reported bug is about a call of empty? on nil and the last report I gave you was about the config

`subnet = unless obj['IPAM']['Config'].empty?`
```
"IPAM": {
            "Driver": "",
            "Options": null,
            "Config": null
        },
```
I guess that this commit is also at risk of having a null Driver instead of an empty object.

davejrt:
You're correct, this was an error on my part.

PR #335, opened 2018-09-19
in puppet 6 undef values are now nil - fixes #334

### LINKED ISSUE #334: Puppet 6: undef values are now nil
Puppet 6 comes with the following change:

```
In versions before Puppet 6.0.0 values from manifests assigned to resource attributes that contained undef values nested in arrays and hashes would use the Ruby symbol :undef to represent those values. When using puppet apply types and providers would see those as :undef or as the string “undef” depending on the implementation of the type. When using a puppet master, the same values were correctly handled. In Puppet 6.0.0 Ruby nil is used consistently for this. (Top level undef values are still encoded as empty string for backwards compatibility). (PUP-9112)
```

This breaks lots of functions, as the all do something like:

```
    if opts['advertise_addr'].to_s != 'undef'
```

duritong:
these changes are puppet 5 & 6 compatible.

### DIFF
--- a/lib/puppet/parser/functions/docker_stack_flags.rb
+++ b/lib/puppet/parser/functions/docker_stack_flags.rb
@@ -8,25 +8,25 @@ module Puppet::Parser::Functions
     opts = args[0] || {}
     flags = []
 
-    if opts['bundle_file'].to_s != 'undef'
+    if opts['bundle_file'] && opts['bundle_file'].to_s != 'undef'
       flags << "--bundle-file '#{opts['bundle_file']}'"
     end
 
-    if opts['compose_files'].to_s != 'undef'
+    if opts['compose_file'] && opts['compose_files'].to_s != 'undef'
       opts['compose_files'].each do |file|
         flags << "--compose-file '#{file}'"
       end
     end
 
-    if opts['resolve_image'].to_s != 'undef'
+    if opts['resolve_image'] && opts['resolve_image'].to_s != 'undef'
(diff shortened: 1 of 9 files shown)


davejrt:
@duritong thanks for your contribution. We have puppet 6 support on our roadmap very shortly, so this will be merged as part of that work very soon.

dsmitty166:
@davejrt any updates on this roadmap?

davejrt:
@dsmitty166 one of the team is currently working on it this sprint

davejrt:
@duritong would you mind rebasing and we'll get this PR merged?

lib/puppet/parser/functions/docker_stack_flags.rb
```diff
@@ -8,23 +8,23 @@ module Puppet::Parser::Functions
...
       flags << "--resolve-image '#{opts['resolve_image']}'"
     end
 
-    if opts['prune'].to_s != 'undef'
+    if opts['prune'] && opts['prune'].to_s != 'undef'
       flags << "--prune '#{opts['prune']}'"
     end
 
-    if opts['with_registry_auth'].to_s != 'undef'
+    if pts['with_registry_auth'] && opts['with_registry_auth'].to_s != 'undef'
```
florindragos: small typo here. should be `if opts['with_registry_auth']`

duritong:
Pushed a rebased version.

davejrt:
closing as this has been fixed in #375

PR #337, opened 2018-09-20
Update readme - Using a proxy on windows

florindragos:


README.md
```diff
@@ -211,6 +211,23 @@ class { 'docker':
 }
 ```
 
+### Proxy on Windows
+To use docker through a proxy on Windows, a System Environment Variable HTTP_PROXY/HTTPS_PROXY must be set. See [Docker Engine on Windows](https://docs.microsoft.com/en-us/virtualization/windowscontainers/manage-docker/configure-docker-daemon#proxy-configuration)
+This can be done using a different puppet module, like puppet-windows_env. After setting the variable, the docker service must be restarted.
```
MWilsonPuppet: @florindragos Hey Florin this is a super picky can you change This can be done using a different puppet module, like puppet-windows_env. To, This can be done using a different puppet module such as the  puppet-windows_env module. Ill then merge for you cheers

florindragos: @MWilsonPuppet Hey Mark, just changed it.

PR #339, opened 2018-09-20
Check for empty IPAM Config

florindragos:
obj['IPAM']['Config'] can be an empty array that can lead to obj['IPAM']['Config'].first.key? throwing an exception:
undefined method `key?' for nil:NilClass


davejrt:
LGTM

PR #341, opened 2018-09-26
CLOUD-2146 cleanup of docker compose acceptance tests

MWilsonPuppet:
This PR is to remove docker-compose tests for compose file versions prior to 3. The tests ran on each version are identical the only difference being the compose file version.Also docker_compose versions being tested were below minimum we support and windows was only supported in version3. Therefore it makes sense to reduce these as they are creating more overhead and not bringing increased value.


davejrt:
LGTM

PR #346, opened 2018-09-27
pinning puppet version to fix failing spec tests

MWilsonPuppet:
This PR pins the puppet version to 5.5.6 as when running with puppet 6 spec tests are failing


florindragos:
👍

PR #355, opened 2018-10-16
WIP: Support for docker config

Damoun:
This PR add a new type `docker_config` to create configs in docker swarm. It could be used as a volumes in docker services/stacks.

Since docker config cannot update the content of the config, this type can just create or remove it.


Damoun:
Just saw that docker secret is supported and could be used instead.

PR #365, opened 2018-10-25
Fixing incorrect variable names in docker_compose/ruby.rb

lowerpuppet:
The variable names weren't changed from "project" to "name" in two places in docker_compose/ruby.rb.  This cause puppet agent failures when certain functions were called when using docker_compose.  A specific example is running docker_compose when the containers already exist.

Per a discussion in the slack channel, I didn't create acceptance tests for this PR since it involves simply correcting two variable names.


davejrt:
LGTM

PR #367, opened 2018-10-30
Create shared start/stop scripts for better extensibility

glorpen:
This PR replaces current docker-runscript with scripts for service start and stop actions - used by systems with and without systemd.
There is not much added functionality for end-user, just singular new parameter - `after_create` in `docker::run`. In my current use case it will allow to easily add connecting container to multiple networks (since `docker create` is now always used).
Additionally this PR opens a way to add handling for named user/group from host system, eg. `--user $(id -u user):$(id -g group)`. (PR's for both will come later)

Currently the first one is possible by overusing `extra_systemd_parameters` parameter and second one is not even possible because of systemd `$()` handling.

Main changes:
- no ExecStartPre/ExecStop(Post) in systemd service
- shared logic for start/stop actions  (shell script), based on existing upstart code
- `docker create` is now always used in init scripts, no `docker run`

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,7 +324,6 @@ This is equivalent to running the  `docker run -d base /bin/sh -c "while true; d
 ```puppet
 docker::run { 'helloworld':
   image            => 'base',
-  detach           => true,
   service_prefix   => 'docker-',
   command          => '/bin/sh -c "while true; do echo hello world; sleep 1; done"',
   ports            => ['4444', '4555'],
--- a/lib/puppet/parser/functions/docker_run_flags.rb
+++ b/lib/puppet/parser/functions/docker_run_flags.rb
@@ -42,10 +42,6 @@ module Puppet::Parser::Functions
       flags << '--privileged'
     end
 
-    if opts['detach']
-      flags << '--detach=true'
-    end
-
     if opts['health_check_cmd'] && opts['health_check_cmd'].to_s != 'undef'
       flags << "--health-cmd='#{opts['health_check_cmd']}'"
     end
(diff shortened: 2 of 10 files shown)


davejrt:
@glorpen can you please resolve the merge conflict here and we can look at getting this merged for the next release.

glorpen:
@davejrt Done, I hope I've catched all run::detach uses.

davejrt:
LGTM

davejrt:
@glorpen looks like this is breaking on systems using upstart. Primarily the -a flag with docker start. Do you want to push a fix, or we'll have to revert the PR so it doesn't break our next release.

glorpen:
@davejrt hmmm, wouldn't reverting f5de8ac0fa705e49690de77356bba7935f3cd9d3 fix it? I didn't noticed it before but `$valid_detach` is still used in inline script and it should result in absence of `-a` flag when not on systemd and now it is undef so I guess always systemd behavior.

davejrt:
I think it makes more sense to leave that out, and fix the init.d script so it runs in detached mode, as was the default behaviour previously. Thoughts?

glorpen:
Done

davejrt:
Do you want to submit the fix as a PR?

glorpen:
If it is required, I'm fine with you just merging/squashing/etc it.

glorpen:
@davejrt not sure if repo silence is caused by holidays or not; let me know if you want me to create that PR after all :)

davejrt:
Happy Holidays, if you could create the PR that would be great

PR #370, opened 2018-11-05
parameterized package_location for osfamily==RedHat

ZyanKLee:
Value of `$docker_ce_source_location` is manufactured in `params.pp` for `osfamily==RedHat` to the exact same URL template.
I suggest using the variable instead, as `$docker_ce_source_location` may be overridden for AmazonLinux for example.


davejrt:
LGTM

PR #373, opened 2018-11-14
Handle compose files with services that have no image

walkamongus:
I ran into this issue when using Puppet to deploy Harbor. Harbor (https://github.com/goharbor/harbor/releases/tag/v1.6.1) comes with multiple compose files and some of them contain services that do not have images associated with them. I needed to build a deep-merged hash of compose services and a array of unique containers from all compose files and move the service vs. container count checking outside of the compose_files.each loop in order to get an accurate comparison of services and their related images.


davejrt:
LGTM

PR #378, opened 2018-11-23
Fix the docker_compose options parameter position

floSoX:
This fix move the potion of docker-compose options from this:
`/usr/local/bin/docker-compose -p --project-directory /path name up -d --remove-orphans`
to this:
`/usr/local/bin/docker-compose -p name --project-directory /path up -d --remove-orphans`


davejrt:
LGTM

kperry42:
I was just perusing the code, and happened to notice ... this PR fixes the args array ordering on one line of the affected file, but **there are five more instances of the exact same bug** later on in the file, which also (still!) need fixing.  All occurrences of "insert(2, ..." should be "insert(3, ".

awegmann:
Encountered the same problem.

PR #383, opened 2018-12-04
docker::run: Support depend_services with full systemd unit names

### LINKED ISSUE #352: Unable to set unit dependencies other then .service
Hi,

When setting a (systemd unit) dependency via `depend_services` the `.service` extension is  added via the erb template but there are different types of units (e.g. .sock, .mount, .device, etc).

So it's currently impossible to create a dependency on anything else than a service.

In my case I need to ensure that an EFS volume is mounted before the docker container is started but adding the unit to `depend_services` results in the follow configuration:

```
[Unit]
Description=Daemon for jenkins
After=docker.service srv.mount.service
Wants=
Requires=docker.service srv.mount.service
```

Which results in a failure during reload:

```
Failed to start docker-jenkins.service: Unit srv.mount.service not found.
```

Reference:

https://www.digitalocean.com/community/tutorials/understanding-systemd-units-and-unit-files

iamjamestl:
To support depending non-service systemd unit types, the docker-run
service template was modified to detect when a full systemd unit name is
passed.  This allows you to depend on targets like 'remote-fs.target' or
'network-online.target' if the container needs it.

Closes #352

### DIFF
--- a/spec/defines/run_spec.rb
+++ b/spec/defines/run_spec.rb
@@ -90,6 +90,16 @@ require 'spec_helper'
           it { should contain_file(initscript).with_content(/After=(.*\s+)?bar.service/) }
           it { should contain_file(initscript).with_content(/Requires=(.*\s+)?foo.service/) }
           it { should contain_file(initscript).with_content(/Requires=(.*\s+)?bar.service/) }
+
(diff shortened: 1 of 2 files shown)


davejrt:

PR #386, opened 2018-12-10
Adds a Usage example for daemon level extra_parameters

mpepping:
The puppetlabs-docker README seems to miss an example on how to add daemon-level parameters. This PR adds:

* Usage for extra_parameters for the Docker daemon
* markdownlint fixes


davejrt:
LGTM

PR #387, opened 2018-12-11
fixing deep merge issue and yaml alias

### LINKED ISSUE #382: docker_compose resource fails to parse YAML aliases
## What you expected to happen?

Using a valid docker-compose YAML file containing aliases should work the same as an equivalent YAML file with the aliases pre-expanded.

## What happened?
Puppet run failed with the following errors:
```
Error: /Stage[main]/Myservice/Docker_compose[myservice]: Could not evaluate: Unknown alias: default-logging
Error: /Stage[main]/Myservice/Docker_compose[myservice]: Failed to call refresh: Unknown alias: default-logging
Error: /Stage[main]/Myservice/Docker_compose[myservice]: Unknown alias: default-logging
```

## How to reproduce it?

Define a docker_compose resource with a compose file similar to one of the examples in https://docs.docker.com/compose/compose-file/#extension-fields

## Anything else we need to know?

The problem seems to be this line of Ruby code: https://github.com/puppetlabs/puppetlabs-docker/blob/fc67e287f2d5ac3858358e34fe7782b83bc05b7f/lib/puppet/provider/docker_compose/ruby.rb#L15

By default, `YAML.safe_load` doesn't handle aliases. According to the documentation at https://ruby-doc.org/stdlib-2.1.0/libdoc/psych/rdoc/Psych.html#method-c-safe_load it should be sufficient to set the `aliases` param like so: `YAML.safe_load(File.read(file), [], [], true)`.

davejrt:
closes #382

### DIFF
--- a/lib/puppet/provider/docker_compose/ruby.rb
+++ b/lib/puppet/provider/docker_compose/ruby.rb
@@ -27,7 +27,7 @@ Puppet::Type.type(:docker_compose).provide(:ruby) do
         compose_services.deep_merge!(compose_file['services'])
       # in compose v1 "version" parameter is not specified
       when nil
-        compose_services = compose_services.deep_merge!(compose_file)
+        compose_services.deep_merge!(compose_file)
       else
         raise(Puppet::Error, "Unsupported docker compose file syntax version \"#{compose_file['version']}\"!")
       end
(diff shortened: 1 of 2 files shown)


florindragos:
👍

PR #389, opened 2018-12-11
Enable PDK compliance

florindragos:


davejrt:
LGTM
@sheenaajay would you mind reviewing also?

PR #391, opened 2018-12-14
Add support for default-addr-pool options to swarm init

sagepe:
Docker API version 1.39 introduced the ability to configure the address pools and CIDR block size used for creating global scope networks. This PR adds support for the `--default-addr-pool` and the associated `--default-addr-pool-mask-length` options to `swarm init`.


davejrt:
LGTM

PR #400, opened 2019-01-02
Fix shared scripts on non systemd systems

glorpen:
This reverts upstream commit "removed $::docker::run::detach parameter".

Problem was introduced with PR #367


davejrt:
LGTM

PR #401, opened 2019-01-06
Move container name sanitisation code to puppet function.

glorpen:
Make sanitisation code available for external modules
so it can be reused when working with scripts and container names.


davejrt:
LGTM

PR #407, opened 2019-01-11
Removing value in the --prune parameter for docker stack deploy

venushka:
Removing value in the --prune parameter for docker stack deploy as the cli doesn't have a value as per the documentation. (see https://docs.docker.com/engine/reference/commandline/stack_deploy/)

With the existing code, if we use the `prune` option as follows, the `docker stack deploy` command fails.

```
  docker::stack { 'proxy': 
    ensure        => present,
    stack_name    => 'proxy',
    compose_files => '/tmp/proxy-compose.yml',
    prune         => true
  }
```

### DIFF
--- a/lib/puppet/parser/functions/docker_stack_flags.rb
+++ b/lib/puppet/parser/functions/docker_stack_flags.rb
@@ -1,11 +1,9 @@
-# frozen_string_literal: true
-
 require 'shellwords'
 #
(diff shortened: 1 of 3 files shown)


davejrt:
LGTM

venushka:
Thanks for merging this in.

PR #421, opened 2019-02-07
Docker::Services: Add 'mounts' parameter

jacksgt:
The --mount parameter for Docker services is the equivalent to
-v/--volume for standalone Docker containers.
This patch allows to directly specify one or multile values for
this argument directly as a Puppet parameter.
This avoids the hassle of having to go through 'extra_params', since
this is a commonly used option.

https://docs.docker.com/storage/volumes/

Please let me know if I forgot to add/edit anything (tests/...)

### DIFF
--- a/README.md
+++ b/README.md
@@ -769,11 +769,12 @@ docker::services {'redis':
     image => 'redis:latest',
     publish => '6379:639',
     replicas => '5',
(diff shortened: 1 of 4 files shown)


davejrt:
LGTM

PR #431, opened 2019-02-15
Allow custom docker_group and service_name values in run.pp

esalberg:
Reopening PR #250 
The default docker_group can be overridden in init.pp, but run.pp only uses the default params values. Also updated service_name the same way.

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -144,8 +144,8 @@ define docker::run(
   }else {
     $docker_command = $docker::params::docker_command
   }
-  $service_name = $docker::params::service_name
-  $docker_group = $docker::params::docker_group
+  $service_name = $docker::service_name
+  $docker_group = $docker::docker_group
 
   if $restart {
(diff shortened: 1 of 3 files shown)


esalberg:
All the checks passed!  I will note that I did not put an explicit include of the main docker class.  Since the rspec tests mock the variables, I don't think testing would cover the case where docker run is used without an include docker somewhere.

Please let me know if you want me to add include docker on the run manifest for this change.

davejrt:
Thanks for getting back to us. 

The only supported functionality is to use the module with the main class included, even if you can use it without so I am happy that this is all the testing covers

davejrt:
looks like this is causing some issues in the acceptance tests, can you push a fix and include the docker class as well as params?

esalberg:
Do you need me to fix the include class in a separate PR?  I have an e-mail notification about needing that change, but I can see my change was merged already (thanks!)...

davejrt:
Hi @esalberg nope all good. Sorry there was an error in our tests that I fixed. Your code was correct :-) I jumped the gun a little

PR #437, opened 2019-02-19
including docker class in all tests of docker::run

davejrt:
fixes errors where docker::service variable is not found.


florindragos:
👍

Issue #439, opened 2019-02-21
Need to manage docker.socket on RHEL (also socket overrides)


esalberg:
## What happened?
As of Docker-ce 18.09.1, the systemd docker.socket is required on RHEL.  This file has a default SocketGroup of docker (even though the docker group is optional per the Docker documentation).  We need a way to override the SocketGroup setting (in /usr/lib/systemd/system/docker.socket).

## How to reproduce it?
On RHEL with Docker 18.09.1+, run "groupdel docker" and try to restart Docker.  it will fail due to the group on /var/run/docker.sock no longer existing (turns numeric).  The initial install / upgrade is fine because the docker group is added by default.

## Anything else we need to know?
I am submitting a PR for this issue tomorrow, which will have more details as to the fix.  I did have a couple of questions I thought I'd ask here, in case this gets seen before the PR:

1) The current docker-systemd-reload-before-service does not actually notify the service to restart.  Is that on purpose, or may I change the before => to a notify => ?

      exec { 'docker-systemd-reload-before-service':
        path        => ['/bin/', '/sbin/', '/usr/bin/', '/usr/sbin/'],
        command     => 'systemctl daemon-reload > /dev/null',
        before      => $_manage_service,
        refreshonly => true,
      }

This is the only breaking change that I'm planning - otherwise, the new socket override settings will default to false overall.

2) The new socket-overrides-rhel.conf.erb file simply has the following:
[Socket]
SocketGroup=<%= @socket_group %>

Any idea if the same would work for Debian?  If so, I can change the overrides file to socket-overrides.conf.erb and specify it as the default file for both.  I don't have Debian so can't check.

davejrt:
closed by #441

Issue #369, opened 2018-11-02
set `$socket_group_default = undef` in params.pp to let it be overridable in hiera


androidkh:
Hi all,

currently there is a parameter $socket_group_default that is used to define $socket_group that in its turn is used in template docker.systemd.erb:

```
<% if @socket_group %> -G <%= @socket_group %><% end -%>
<% if @dns %><% @dns_array.each do |address| %> --dns <%= address %><% end %><% end -%>
```

By default on centos7 it resulted in adding "-G docker" to OPTIONS.
I have existing infrastructure that has no -G specified at all (and thus it's a root by default).
I found no way to override it in hiera. Even when I specified empty string -  <% if @socket_group %> was true and I just received `... -G --dns...`
When I tried to specify undef for this param - then it took default value 'docker' from params.pp

Is it possible to modify the code to be able to set $socket_group to undef?
I found the only way - not to assign any value in param.pp. In this case socket_group can by set in hiera. Any other ways are welcome as well.
Thanks, 
docker module version: 3.0.0

mojibake-umd:
This is a dup of #321 in my opinion.
So current situation, broke.

esalberg:
Following up on this - we used to have the same issue, but with the updated socket management for RHEL in Docker 18.09.1, setting socket_group to undef wouldn't work anyway.  We need to set it explicitly to root and add a /etc/systemd/system/docker.socket.d/socket-overrides.conf file to override the new default in /usr/lib/systemd/system/docker.socket (SocketGroup=docker).  It does mean that you'll have "-G root" in your OPTIONS line.

Which OS are you using?  I can only test with RHEL, but the ability is there to specify an override template for other types of Linux as well (disabled by default across the board).  I'll be putting in a PR shortly and will link.

androidkh:
Idea was to allow "not to specify" -G option at all  as docker can use default value when no -G flag.

esalberg:
In my PR, that is done by setting:

docker::socket_group: false

Since all the templates have <% if @socket_group %> wrappers around the socket_group related entries (including the -G option), it works fine in my testing.

davejrt:
closed by #441

PR #441, opened 2019-02-21
Add docker socket group overrides for systemd (service and socket)

esalberg:
This patch allows systemd overrides for the /usr/lib/systemd/system/docker.socket file (new for RHEL as of 18.09.1).  It also allows legacy configs to override setting the socket group in /etc/sysconfig/docker (i.e. no -G setting).

It also causes any changes made to the service or socket overrides.conf to notify the service to restart (before, systemd daemon-reload would occur, but the service was just ordered, not notified).

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,21 @@ class { 'docker':
 }
 ```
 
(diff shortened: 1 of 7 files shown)


esalberg:
Please note: I am not able to test Debian, but I figured that socket file was generic enough that the same overrides template might work.  If not, I can remove that default and change the template file not to specify RHEL.  (Since the override is disabled by default, I figured I might as well add the filename.)

Ref:
#439 
#369

davejrt:
LGTM

PR #446, opened 2019-02-28
Add ability to override After for docker.service

esalberg:
There is a recent patch for docker-ce related to systemd ordering that has not made it to a stable release yet.  In order to apply the workaround, it's necessary to be able to override (add another service) to the After= line in the [Unit] section of docker.service.

Ref: https://github.com/docker/docker-ce-packaging/pull/290

Note that I added the option for Debian too but was not able to test it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,8 @@ class { 'docker':
 }
 ```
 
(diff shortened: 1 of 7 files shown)


davejrt:
LGTM

@esalberg you can run against ubuntu in this repo by running `rake acceptance:vagrant:ubuntu-1604-x64`

PR #447, opened 2019-03-02
docker::services: Fix using multiple published ports

jacksgt:
While the class docker::services already allowed arrays to be
specified for `publish` ports, it didn't handle this correctly since
this array was simply converted to a string and passed onto Docker,
like so:

    --publish '["80:80", "443:443"]'

To correctly publish multiple ports, the `--publish` parameter has to
be called specified for each port.

    --publish '80:80' --publish '443:443'

This patch also introduces a new spec test to check this behavior

### DIFF
--- a/README.md
+++ b/README.md
@@ -799,7 +799,7 @@ docker::services {'redis':
   }
 ```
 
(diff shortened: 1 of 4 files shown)


davejrt:
LGTM

PR #448, opened 2019-03-06
Add ability to set service_after_override to false

esalberg:
Need to be able to override a value in hiera so that no value is set.


davejrt:
LGTM

PR #450, opened 2019-03-08
Docker::Services: Add networks parameter for swarm services

jacksgt:
Allows attaching the service to one or multiple networks

Also adds spec tests to check the functionality


davejrt:
LGTM

PR #452, opened 2019-03-09
Docker::Services:: fix command parameter used with an array

jacksgt:
The module previously allowed specifying the command parameter to
docker::services as an array, but did not correctly parse it as
such (output with square brackets and quotations marks to docker).
This patch fixes this behavior.


davejrt:
LGTM

PR #462, opened 2019-03-21
Update docker compose version to latest

tarjei:
I havent tested it extensively so if there is a special reason you are using an old version then this PR is moot.

https://github.com/docker/compose/releases


davejrt:
If someone is using compose without pinning the version this will introduce a breaking change.

lionce:
Hello @tarjei ,

Thank you for your PR! By changing the value of docker compose in module's manifest, the existing functionality may break. In order to use the latest version, you should specify it in your local manifest! Thanks again for your contribution! I'll close this PR!

Cheers!

PR #465, opened 2019-03-22
Fix docker::image to not run images

hugotanure:
According the documentation, the resource docker::image should just to pull an image, not properly try to run.

```
To install a Docker image, add the docker::image defined type to the manifest file:

docker::image { 'base': }
The code above is equivalent to running the docker pull base command. However, it removes the default five-minute execution timeout.
```

### DIFF
--- a/manifests/image.pp
+++ b/manifests/image.pp
@@ -153,5 +153,5 @@ define docker::image(
     }
   }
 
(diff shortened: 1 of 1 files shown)


davejrt:
LGTM

PR #451, opened 2019-03-08
adding in capability to attach multiple networks to container

davejrt:
fixes #445

PR #470, opened 2019-04-08
Fix container running check to work for windows hosts

florindragos:


manifests/run.pp
```diff
@@ -320,7 +320,7 @@
       } else {
         exec { "start ${title} with docker":
           command     => "${docker_command} start ${sanitised_title}",
-          onlyif      => "${docker_command} inspect ${sanitised_title} -f \"{{ .State.Running }}\" | grep false",
+          onlyif      => "${docker_command} inspect ${sanitised_title} -f \"{{ if (.State.Running) }} {{ nil }}{{ end }}\"",
```
tphoney: so it is the same command in both places. the check is different, ie onlyif vs unless. Am i correct in that summation ?

glennsarti:
* Could you also update the README as it says Windows isn't supported
https://github.com/puppetlabs/puppetlabs-docker/blob/master/README.md#limitations

* What about Windows Server 2019 - Missing from Metadata

* This module isn't tested on Windows (at least no Appveyor, only Travis CI) so it's difficult to know if this even works.

manifests/run.pp
florindragos: was supposed to be a revert from https://github.com/puppetlabs/puppetlabs-docker/pull/451 but I just realised this does not work also. that's why dave changed it to use grep but we don't have that on windows so i'll have to find a command that works on all OSs...

florindragos:
@glennsarti Good catch about the Readme
We haven't tested on Windows 2019 and I don't have any knowledge of someone requesting it (yet) but we should certainly add support for it.

glennsarti:
Given the functionality you're leveraging, you can make a really good argument that testing on 2016 would satisfy the 2019 requirement i.e. it's already mostly tested and not worth the effort to to acceptance testing on 2019.

tphoney:
Great work @florindragos  ! 👍

PR #471, opened 2019-04-08
Fix publish flag being erroneously added to docker service commands

twistedduck:
Currently adding a `docker::services` resource without publishing any ports in puppet6 fails as the command is run with the invalid flag: `--publish ''`.


florindragos:
👍 nice catch

PR #480, opened 2019-05-06
Update docker volume documentation

benningm:
The documentation of

* docker_volume type
* docker::volumes puppet class

in README.md seems to be outdated and incorrect.

This tries to fix the docs and adds an example for
using an NFS volume.


carabasdaniel:
@benningm Thank you for your contribution.

PR #484, opened 2019-05-16
(MODULES-9029) Update pdk-template to 46a1deb

eimlav:


eimlav:
Adhoc ran successfully:

Screen Shot 2019-05-17 at 11 48 25

PR #487, opened 2019-05-22
(FM-8055) Add RedHat 8 support

eimlav:


eimlav:
Closing as Docker (engine/CLI) is not supported on RHEL8.

PR #490, opened 2019-05-30
(MODULES-9193) Revert part of MODULES-9177

eimlav:
The changes introduced in MODULES-9177 modified the behaviour of docker
installs using older versions. The first part of the modified regex ensured
that the correct repository was used. As such I have reverted it but
left the rest of the changes to ensure that non 17.y.z versions use the
correct repo.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -594,7 +594,7 @@ class docker(
     }
   }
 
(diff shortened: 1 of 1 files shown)


eimlav:
Adhoc ran successfully:

Screen Shot 2019-05-30 at 16 49 35

PR #497, opened 2019-06-14
Fix #239 local_user permission denied

### LINKED ISSUE #239: Local user and idempotent operation with registry logins to same host, different users
# What happened?
My registry logins are overwritten by logins to registries on the same server.

# What did I expect?
I expected to be able to log in to the same registry with different users.

When using docker::registry to log in to a registry, it appears that it will overwrite a previous login if the server name (and local user) matches. The docker::registry resources run first, which causes some registries to be overwritten.

In the code,
```
# [*local_user*]
#   The local user to log in as. Docker will store credentials in this
#   users home directory
```

It appears that the receipts are stored in /root, though? This seems fine for my use-case, but it seems like having the receipt generated with a resource name or similar could be beneficial. It also seems to be a bug in the code comment since it seems fixed to store those receipts in /root, but that's nonsensical as the $local_user wouldn't have access to write to /root

thde:
This is an approach to fixing the `local_user` permission issues (fixes #239). Instead of writing to `/root` the users home directory is used.

### DIFF
--- a/lib/facter/docker.rb
+++ b/lib/facter/docker.rb
@@ -33,12 +33,12 @@ Facter.add(:docker_user_temp_path) do
 end
 
 Facter.add(:docker_home_dirs) do
-  confine :kernel => 'Linux'
+  confine kernel: 'Linux'
   setcode do
     home_dirs = {}
-    Etc.passwd { |user|
+    Etc.passwd do |user|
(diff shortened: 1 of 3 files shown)


florindragos:
Thanks for this fix!
Acceptance tests are all green so I will merge this! 👍

PR #512, opened 2019-07-18
Add ability to override checks when using restart policy

esalberg:
Also allows overriding the container name - this is useful when a container spawns containers with different names (e.g. Rancher nodes, which start with rancher-worker and spawns a number of containers including kubelet).

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -62,7 +62,6 @@
 # Default: undef
 #
 # [*net*]
-#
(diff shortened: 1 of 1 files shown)


esalberg:
Please don't merge this yet.  The container_name change may not do what I need it to do.

esalberg:
I believe I fixed the code, but I'm still testing, so I'm closing the change and will resubmit.

Issue #510, opened 2019-07-14
Installing plugin fails if alias is specified


Fuco1:
## What you expected to happen?

I expect the plugin installation, if it is already installed, to not trigger an error.

## What happened?

Puppet agent run fails with a message that the plugin is already installed.

## How to reproduce it?

I install a plugin with 

``` puppet
docker::plugin {'docker4x/cloudstor:18.09.2-ce-azure1':
    ensure   => 'present',
    plugin_alias => 'cloudstor:azure',  ## Notice the alias
    settings => [
      'CLOUD_PLATFORM=AZURE',
    ],
  }
```

An alias is specified.  This will make the output of `docker plugin ls` print something like:

```
ID                  NAME                DESCRIPTION                       ENABLED
e33c236ec53b        cloudstor:azure     cloud storage plugin for Docker   true
```

This package uses `$package_name` to grep this output to see if the plugin is specified.  This will never return true because the plugin name is not used in the output.  Puppet will then try to reinstall the plugin which results into

```
Notice: /Stage[main]/Profile::Docker/Docker::Plugin[docker4x/cloudstor:18.09.2-ce-azure1]/Exec[plugin install docker4x/cloudstor:18.09.2-ce-azure1]/returns: Error response from daemon: plugin cloudstor:azure already exists
```

## Anything else we need to know?

Possible solution would be to (I think) change the test 

``` puppet
$unless_install = "${docker_command} ls | grep -w ${plugin_name}"
```

to use the `$plugin_alias` if specified.

## Versions:

```
matus@worker-0:/mnt/cloudstor/sharedvol1$ puppet --version
6.4.2
matus@worker-0:/mnt/cloudstor/sharedvol1$ docker version
Client:
 Version:           18.09.6
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        481bc77
 Built:             Sat May  4 02:35:57 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.6
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.8
  Git commit:       481bc77
  Built:            Sat May  4 01:59:36 2019
  OS/Arch:          linux/amd64
  Experimental:     false
matus@worker-0:/mnt/cloudstor/sharedvol1$ facter os
{
  architecture => "amd64",
  distro => {
    codename => "bionic",
    description => "Ubuntu 18.04.2 LTS",
[...]

PR #514, opened 2019-07-24
Fix aliased plugin names

koshatul:
#### What change does this introduce?

- Adds the option `plugin_alias` to `docker_plugin_enable_flags`.
- Adds `--format='{{.PluginReference}}'` to the plugin ls commands.

#### Why make this change?

Current code assumes that plugin_name is the same as plugin_alias, even though plugin_alias is parameter.

#### What issues does this relate to?

- #510


sheenaajay:
Thanks for submitting the PR @koshatul .Could you please add tests too for them.

koshatul:
Added RSpec tests.

Also the [CONTRIBUTING.md](CONTRIBUTING.md) file has some dud links.

koshatul:
Also fixed the RuboCop warnings on the spec files I modified.

PR #516, opened 2019-07-26
Fixing error:

darshannnn:
Fixing the following error: 
Filepath: /etc/puppetlabs/code/environments/production/modules/docker/templates/etc/systemd/system/docker-run.erb
  Line: 3
  Detail: undefined method `+' for nil:NilClass
======
replaced direct concatenation of @service_name with a check for nil and then concatination


sheenaajay:
Thanks for submitting the PR @darshannnn .

sheenaajay:
Screen Shot 2019-08-06 at 09 33 59

PR #517, opened 2019-08-08
(MODULES-9696) remove docker_home_dirs fact

vchepkov:
enumerating existing users is insecure and resource consuming
user with non-standard home needs to path it explicitly when creating docker::registry resource

### DIFF
--- a/manifests/registry.pp
+++ b/manifests/registry.pp
@@ -40,6 +40,7 @@ define docker::registry(
   Optional[String] $pass_hash                          = undef,
   Optional[String] $email                              = undef,
   Optional[String] $local_user                         = 'root',
+  Optional[String] $local_user_home                    = undef,
   Optional[String] $version                            = $docker::version,
   Optional[Boolean] $receipt                           = true,
 ) {
@@ -61,7 +62,16 @@ define docker::registry(
     $exec_provider = undef
     $password_env = "\${password}"
     $exec_user = $local_user
-    $local_user_home = $facts['docker_home_dirs'][$local_user]
+
+    if $local_user_home {
+      $_local_user_home = $local_user_home
+    } else {
+      # set sensible default
+      $_local_user_home = $local_user == 'root' ? {
+        true    => '/root',
+        default => "/home/${local_user}",
+      }
+    }
   }
 
   if $ensure == 'present' {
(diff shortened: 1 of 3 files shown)


manifests/registry.pp
```diff
@@ -40,6 +40,7 @@
   Optional[String] $pass_hash                          = undef,
   Optional[String] $email                              = undef,
   Optional[String] $local_user                         = 'root',
+  Optional[String] $local_user_home                    = undef,
```
michaeltlombardi: The inclusion of a new parameter leads me to think that there should be tests in the spec/defines/registry_spec.rb to verify the behavior when this parameter is/is not called, as with username in the linked example;

https://github.com/puppetlabs/puppetlabs-docker/blob/master/spec/defines/registry_spec.rb#L28-L38

RandomNoun7:
Hello @vchepkov,
As my colleague Michael has pointed out, the unit tests will need to be updated to reflect the new parameter added to the defined type.

Please also note however, that your change assumes that the home directory for the `$local_user` is going to be `/home/${local_user}`, but this is not always going to be a safe assumption. Users could possibly place their home directory somewhere else.

We agree that the implementation of the facter fact that was previously used to find the user's home directory had issues, but you still need to be able handle cases where this directory is in a non-standard location.

We think it would improve the quality and functionality of this PR if you were to use the `etc` gem as an internal function instead of a facter fact, and instead of iterating over the entire `Etc.passwd` collection, use the `getpwuid(uid)` function to query specifically for the `$local_user` home directory, and use that information to build the path on line 72 of `registry.pp`. A good way to go might be to implement the function so that it fully builds the path for you, and it's return is a string path. That way you can also modify that function in the future if needed, to account for more complex edge cases, and the `registry.pp` file that calls the function doesn't need to change.

You can use the [modern API for writing custom functions](https://puppet.com/docs/puppet/5.5/functions_ruby_overview.html) to write the function in Ruby so you can require the gem

Thanks, and let us know if you have any questions or need additional guidance.

vchepkov:
That was the idea, if you have a non-standard home you can pass it as an argument.
Correct me if I am wrong, but functions are executed on the master, not on the agent

vchepkov:
I will try to add a test though, not my strongest suite

RandomNoun7:
@vchepkov Sorry it took a little while to get back to you. I think I see a little bit better now how you have that parameter set up, so that part looks good. Any progress on writing tests?

michaeltlombardi:
Since this hasn't seen any activity in about fivemonths, I'm going to go ahead and close this pull request for the time being. Please re-open this pull request once the next actions are addressed, new information is available, or you have a question related to this pull request. We've become aware of difficulties re-opening pull requests, in the event you cannot please mention `puppetlabs/modules`  with an @ sign in front and we'll re-open this pull request.

Closing the pull request doesn't mean we don't consider this change valuable, just that there are things that need to be addressed before it can be merged. If you have any questions or concerns, please don't hesitate to ping us in slack.puppet.com.

PR #520, opened 2019-08-24
:bug: Fix wrong service detach handling

khaefeli:
the service_name always needs to be first argument in the flags array, otherwise the `service create` cmd results in a wrong order / syntax:

https://github.com/puppetlabs/puppetlabs-docker/blob/master/manifests/services.pp#L150

besides that, checking "if not true" results in the opposite behaviour as expected.
which means, setting the puppet param detach to true, results in not setting it on the cmd :)


florindragos:
👍

Issue #542, opened 2019-10-28
Allow matching of docker channel version on Windows.


Dan70402:
## Use Case
Set the version family of the 'Docker' package to use from the DockerMsftProvider.  The provider endpoint may rotate out specific patch level version.  For example,  I had set `19.03.1` as a specific version to pin our servers to, but it has since been rotated out from the provider json metadata in favor of a recent patch version.  I'd like to use the `19.03` *channel* instead.  This will ensure I get a 19.03.x version of docker which will be compatible for my use in swarm clusters where master and workers should be on the same minor version.

## Describe the Solution You Would Like
Change the behavior of the `check_powershell_provider.ps1` to match on version string compares instead of a 'not equal'.  This will maintain existing behavior for specific version and allow me to use the `channel` 19.03.  As I provision new servers they will be guaranteed to be on the same major.minor version.

## Describe Alternatives You've Considered
1. Hosting the package myself 
2. Finding a way to 'flip' the `manage_service` attribute to the service is only managed on initial install

## Additional Context
I have create a PR with the proposed improvement:

https://github.com/puppetlabs/puppetlabs-docker/pull/543

PR #543, opened 2019-10-28
feature/match_docker_windows_version

Dan70402:
On Windows `-notmatch` instead of `-ne` on version compared.  This will allow the use of version channels from DockerMsftProvier.  

PR for proposed solution at:

https://github.com/puppetlabs/puppetlabs-docker/issues/542


lionce:
Hello @Dan70402 ,

This change looks good to me. 👍  Thank you for this contribution!

_Some extra details here
-notmatch | Returns true when string does not match
-ne | Return a value of TRUE or the matches when one or more of the input values is identical to the specified pattern. The entire pattern must match an entire value._

Cheers!

PR #544, opened 2019-10-30
Allow defining the name of the docker-compose symlink

gtufte:
I bumped into a problem with docker-compose where every docker-compose command failed with:

```
docker compose: Error while loading shared libraries: libz.so.1: failed to map segment from shared object: Operation not permitted
```
 
Found out it's related to docker-composes use of the environment variable `TMPDIR`, usually pointing to /tmp or /var/tmp, and that docker-compose expects to be able to run scripts from this path. Being able to run scripts from these locations is not a security best practice, so I made a wrapper script installed at `/usr/local/bin/docker-compose` where I set TMPDIR to a safer location, before running the actual docker-compose binary. This wrapper script now conflicts with the default symlink file created by this fabulous module, so this PR adds the ability to give this symlink another name.

The issues I'm referring to is also described in more detail on this stackoverflow post:
https://stackoverflow.com/questions/57796839/docker-compose-error-while-loading-shared-libraries-libz-so-1-failed-to-map-s

### DIFF
--- a/manifests/compose.pp
+++ b/manifests/compose.pp
@@ -17,6 +17,12 @@
 #   The path where to install Docker Compose.
 #   Defaults to the value set in $docker::params::compose_install_path
 #
+# [*symlink_name*]
+#   The name of the symlink created pointing to the actual docker-compose binary
+#   This allows use of own docker-compose wrapper scripts for the times it's
+#   necessary to set certain things before running the docker-compose binary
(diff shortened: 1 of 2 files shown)


florindragos:
Hello @gtufte,
Thanks for your contribution!
Fix looks good. I'll run the acceptance tests and merge it if everything is ok.

Issue #552, opened 2019-11-18
Puppet run does not show an error whe try to pull an image:tag that does not exists, using ensure => latest


rafaelcarv:
## Describe the Bug
Since the version 3.6.0 of puppetlabs-docker, when if I use a wrong tag for a image and **ensure => 'latest'**, the puppet run will not show an error. 

Would be good to show a error message in this case, because would be easier to know that the **image:tag** user has a typo, or was not created.

## Expected Behavior
It would be expected for the puppet run show an error, like it did in the version 3.5.0 or when using **ensure => present**

The expected error message would be:
`Notice: /Stage[main]/Profiles::Cluster/Profiles::Cluster::Docker[mongo-express]/Docker::Image[mongo-express:oldold]/Exec[/usr/local/bin/update_docker_image.sh mongo-express:oldold]/returns: Error response from daemon: manifest for mongo-express:oldold not found
Notice: /Stage[main]/Profiles::Cluster/Profiles::Cluster::Docker[mongo-express]/Docker::Image[mongo-express:oldold]/Exec[/usr/local/bin/update_docker_image.sh mongo-express:oldold]/returns: Docker image mongo-express:oldold failed to pull!
Error: '/usr/local/bin/update_docker_image.sh mongo-express:oldold' returned 3 instead of one of [0,2]
Error: /Stage[main]/Profiles::Cluster/Profiles::Cluster::Docker[mongo-express]/Docker::Image[mongo-express:oldold]/Exec[/usr/local/bin/update_docker_image.sh mongo-express:oldold]/returns: change from notrun to 0 2 failed: '/usr/local/bin/update_docker_image.sh mongo-express:oldold' returned 3 instead of one of [0,2]
Notice: /Stage[main]/Profiles::Cluster/Profiles::Cluster::Docker[mongo-express]/Docker::Image[mongo-express:oldold]/Exec[/usr/local/bin/update_docker_image.sh mongo-express:oldold]/returns: Error response from daemon: manifest for mongo-express:oldold not found
Notice: /Stage[main]/Profiles::Cluster/Profiles::Cluster::Docker[mongo-express]/Docker::Image[mongo-express:oldold]/Exec[/usr/local/bin/update_docker_image.sh mongo-express:oldold]/returns: Docker image mongo-express:oldold failed to pull!
[...]

PR #553, opened 2019-11-18
Fix error does not show when image:tag does not exists (#552)

rafaelcarv:
Using an exec resource instead of the onlyif will execute the script in a resource, not in a condition. This will make an error to be shown when an image:tag does not exists, not just not making an echo to be shown when the script fails.
Doing the execution of the script this way will be easier to track when a image:tag has a typo or was not created yet.

### DIFF
--- a/manifests/image.pp
+++ b/manifests/image.pp
@@ -144,11 +144,19 @@ define docker::image(
     notify { "Check if image ${image_arg} is in-sync":
       noop      => false,
     }
+    ~> exec { $image_install:
+      environment => $exec_environment,
+      path        => $exec_path,
+      timeout     => $exec_timeout,
+      returns     => ['0', '2'],
+      require     => File[$update_docker_image_path],
+      provider    => $exec_provider,
+      logoutput   => true,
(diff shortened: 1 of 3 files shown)


florindragos:
Hey @rafaelcarv 
Fix looks good. Ran the acceptance tests and they are green!
Thanks for your contribution!

electrofelix:
what was the point in leaving the second `exec` in place? As opposed to replacing it entirely?

manifests/image.pp
```diff
@@ -144,11 +144,19 @@
...
       noop      => false,
     }
+    ~> exec { $image_install:
+      environment => $exec_environment,
+      path        => $exec_path,
+      timeout     => $exec_timeout,
+      returns     => ['0', '2'],
+      require     => File[$update_docker_image_path],
+      provider    => $exec_provider,
+      logoutput   => true,
```
electrofelix: This probably needs the `refreshonly` metaparam set to true to ensure noop behaviour works as before.

PR #561, opened 2019-12-19
Minimal changes to work with podman-docker

seriv:
Podman does not have a service, so 'docker.service' moved to default arguments enabling docker-run to use empty 'depend_services' thus getting rid of it. And facter library modified to reflect "Version" in different place.


seriv:
Will open with fixed systemd unit template.

PR #580, opened 2020-02-26
Add after_start and after_stop options to docker::run define

jantman:
This pull request adds ``after_start`` and ``after_stop`` parameters to the ``docker::run`` define, working similarly to the existing ``before_start`` and ``before_stop`` options (i.e. these execute a command _after_ the container is started or stopped).

PR includes unit tests. I haven't added these to the acceptance tests as the existing ``before_`` variants weren't being tested there, but these are relatively simple template changes and we're currently using this code.


sheenaajay:
Thanks, @jantman for submitting the PR. Great work.

PR #585, opened 2020-03-06
[IAC-291] Convert acceptance tests to Litmus

carabasdaniel:


spec/spec_helper_acceptance_local.rb
```diff
@@ -0,0 +1,190 @@
...
+    image: winamd64/hello-seattle:nanoserver
+      EOS
+    docker_stack_override_windows2016 = <<-EOS
+version: "3"
+services:
+  compose_test:
+    image: winamd64/hello-seattle:nanoserver-sac2016
+      EOS
+    if os[:family] == 'windows'
+      create_remote_file(host, '/tmp/docker-compose-v3.yml', docker_compose_content_v3_windows)
```
cmccrisken-puppet: Forgive me if you already explained this - I seem to recall a conversation about it. 🙇 

Should we be creating files in `/tmp` on Windows hosts. My memory is fuzzy but you may have mentioned in stand-up that this is not being invoked for now?

spec/spec_helper_acceptance_local.rb
```diff
@@ -0,0 +1,190 @@
...
+
+    next unless os[:family] == 'windows'
+    result = run_shell("ipconfig | findstr /i 'ipv4'")
+    raise 'Could not retrieve ip address for Windows box' if result.exit_code != 0
+    ip = result.stdout.split("\n")[0].split(':')[1].strip
+    retry_on_error_matching(60, 5, %r{connection failure running}) do
+      @windows_ip = ip
+    end
+    apply_manifest("class { 'docker': docker_ee => true, extra_parameters => '\"insecure-registries\": [ \"#{@windows_ip}:5000\" ]' }", catch_failures: true)
+    docker_path = '/cygdrive/c/Program Files/Docker'
```
cmccrisken-puppet: Do we still have a dependency on Cygwin?

carabasdaniel: we should not

spec/spec_helper_acceptance_local.rb
carabasdaniel: the paths should be translated

PR #591, opened 2020-03-31
Allow module to work on SLES 12

npwalker:
Prior to this commit, if you set acknowledge_unsupported_os = true
you'd still fail on SLES 12 because the docker::run define
doesn't know that SLES uses systemd and drops down to failing

After this commit, we copy the default section in params for
suse and set the service_provider to systemd so the run define
will work.

The user still needs to manually install docker on SLES 12 but this
allows managing the service etc... to work.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -278,6 +278,46 @@ class docker::params {
       $service_hasrestart                  = undef
       $detach_service_in_init              = true
     }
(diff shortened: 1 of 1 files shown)


DavidS:
Merging without further tests or support to accommodate the use on unsupported SLES.

Issue #485, opened 2019-05-16
Exec auth runs with every catalog apply


catsem:
## What you expected to happen?
When private docker registry configured, Exec auth should run once not with every catalog apply.

## What happened?

``
root@moon:~# puppet agent -t
Info: Using configured environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Info: Caching catalog for moon.milkyway
Info: Applying configuration version '1558011091'
Notice: /Stage[main]/Docker::Registry_auth/Docker::Registry[foo.bar:5000]/File[/root/registry-auth-puppet_receipt_foo.bar:5000_root]/ensure: defined content as '{md5}3232d3ba033fdfac7de67807b5677e23'
Info: /Stage[main]/Docker::Registry_auth/Docker::Registry[foo.bar:5000]/File[/root/registry-auth-puppet_receipt_foo.bar:5000_root]: Scheduling refresh of Exec[foo.bar:5000 auth]
Notice: /Stage[main]/Docker::Registry_auth/Docker::Registry[foo.bar:5000]/Exec[foo.bar:5000 auth]: Triggered 'refresh' from 1 events
Notice: Applied catalog in 1.47 seconds
``

## How to reproduce it?
Add docker module to your env and config private registry like this (using hiera):

docker::extra_parameters:
    - --userns-remap="default"    
docker::registry_auth::registries:
  'foo.bar:5000':
    username: 'user'
    password: 'password'

## Versions:

```
$ puppet --version
4.10.9

$ docker version
Client:
 Version:       17.12.0-ce
 API version:   1.35
 Go version:    go1.9.2
 Git commit:    c97c6d6
 Built: Wed Dec 27 20:11:19 2017
 OS/Arch:       linux/amd64

Server:
 Engine:
  Version:      17.12.0-ce
  API version:  1.35 (minimum version 1.12)
  Go version:   go1.9.2
  Git commit:   c97c6d6
  Built:        Wed Dec 27 20:09:53 2017
  OS/Arch:      linux/amd64
  Experimental: false

$ facter os
{
  architecture => "amd64",
  distro => {
    codename => "xenial",
    description => "Ubuntu 16.04.6 LTS",
    id => "Ubuntu",
    release => {
      full => "16.04",
      major => "16.04"
    }
  },
  family => "Debian",
  hardware => "x86_64",
  name => "Ubuntu",
  release => {
    full => "16.04",
    major => "16.04"
[...]

PR #610, opened 2020-05-12
[MODULES-10629] Throw error when docker login fails

carabasdaniel:
This PR should highlight if there are any issues when docker login is used, as a result this might solve issues such as https://github.com/puppetlabs/puppetlabs-docker/issues/485 where the login seems to silently fail.

### DIFF
--- a/manifests/registry.pp
+++ b/manifests/registry.pp
@@ -113,7 +113,7 @@ define docker::registry(
       # server may be an URI, which can contain /
       $server_strip  = regsubst($server, '[/:]', '_', 'G')
       $passfile      = "${::docker_user_temp_path}/registry-auth-puppet_receipt_${server_strip}_${local_user}"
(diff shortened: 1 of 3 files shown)


manifests/registry.pp
```diff
@@ -113,7 +113,7 @@
       # server may be an URI, which can contain /
       $server_strip  = regsubst($server, '[/:]', '_', 'G')
       $passfile      = "${::docker_user_temp_path}/registry-auth-puppet_receipt_${server_strip}_${local_user}"
-      $_auth_command = "if (-not (${auth_cmd})) { Remove-Item -Path ${passfile} -Force -Recurse -EA SilentlyContinue; exit 0 } else { exit 0 }" # lint:ignore:140chars
+      $_auth_command = "if (-not (${auth_cmd})) { Remove-Item -Path ${passfile} -Force -Recurse -EA SilentlyContinue; exit 1 } else { exit 1 }" # lint:ignore:140chars
```
DavidS, line 116:
```suggestion
      $_auth_command = "if (-not (${auth_cmd})) { Remove-Item -Path ${passfile} -Force -Recurse -EA SilentlyContinue; exit 1 } else { exit 0 }" # lint:ignore:140chars
```

my understanding is the second `exit` is the successful case?!?

Issue #575, opened 2020-02-01
Unsupported docker compose file syntax version "3.7"


tobiashochguertel:
I receive with a docker-compose file which has version Number 3.7 setuped, an error.

```
 STDERR:
    Error: /Stage[main]/Project_status_monitor::Install/Docker_compose[Docker-Compose[Status-Monitor]]: Could not evaluate: Unsupported docker compose file syntax version "3.7"!
```

## Environment
mod 'puppetlabs-docker', '3.9.1'

```
version: "3.7"

services:
  nginx-proxy:
    image: jwilder/nginx-proxy
    ports:
      - "80:80"
    volumes:
      - /var/run/docker.sock:/tmp/docker.sock:ro
      - "nginx_proxy_data_vhostd:/etc/nginx/vhost.d"
      - "nginx_proxy_data_confd:/etc/nginx/conf.d"
    networks:
      - default
      - frontend
```

The Bug is located in the following file: https://github.com/puppetlabs/puppetlabs-docker/blob/master/lib/puppet/provider/docker_compose/ruby.rb

vegaaz:
Go to your host and do "docker-compose --version" , when it is unix/linux it will be 1.9.0 which has no support for 3.7 ( I think so).
You can set docker-compose version:
  class { 'docker::compose':
    version => '1.21.0'
  }

PR #620, opened 2020-06-02
Allow all 3.x docker-compose minor versions

runejuhl:
No major changes affecting this module compared to the previous max
compatibility with docker-compose 3.6; see
https://docs.docker.com/compose/compose-file/compose-versioning/#version-37

Fixes https://github.com/puppetlabs/puppetlabs-docker/issues/575


daianamezdrea:
Hi @runejuhl, this LGTM and the tests are green, thank you for your contribution! Cheers!

PR #622, opened 2020-06-11
Blocking ordering between non-Windows service stops

ALTinners:
Windows based services stopping will block until the docker container stops (ie: if `docker::run::stop_wait_time` has been set to !=0)

`service` based stops do not - this adds that dependency in so that containers are not attempted to be removed until the service has stopped


adrianiurca:
Hi @ALTinners,
Thank you for your contribution.

PR #631, opened 2020-07-08
Fixing the fix 'Fix the docker_compose options parameter position #378'

awegmann:
The fix done in 2018 was not complete. There were five more code lines which inserted the optional parameters at the wrong position.


daianamezdrea:
Hi @awegmann, thank you for your contribution! Cheers!

PR #632, opened 2020-07-08
(MODULES-10691) - Add root_dir in daemon.json

daianamezdrea:


daianamezdrea:
This solves [10691](https://tickets.puppetlabs.com/browse/MODULES-10691)

spec/acceptance/docker_spec.rb
```diff
@@ -43,12 +43,25 @@
...
     end
 
+    context 'Checking root_dir value' do
+      let(:pp) do
+        "class { 'docker': #{docker_args}}"
+      end
+
+      it 'is good' do
+        apply_manifest(pp, catch_failures: true)
+        run_shell('cat C:/ProgramData/docker/config/daemon.json') do |r|
```
sanfrancrisko, line 59: There's helpers that can verify the content of the file in `puppet-rspec` - see [here](https://github.com/puppetlabs/puppetlabs-apache/blob/f8af65497f7198eac5e6eedd4bae529f561f2da4/spec/acceptance/mod_ldap_spec.rb#L15-L17)

...although this is fine 👍 

One thing I might suggest is verifying the value of `data-root` too?

spec/acceptance/docker_spec.rb
```diff
@@ -9,12 +9,12 @@
   raise 'Could not retrieve ip address for Windows box' if result.exit_code != 0
   ip = result.stdout.split("\n")[0].split(':')[1].strip
   @windows_ip = ip
-  docker_arg = "docker_ee => true, extra_parameters => '\"insecure-registries\": [ \"#{@windows_ip}:5000\" ]'"
+  docker_args = "docker_ee => true, extra_parameters => '\"insecure-registries\": [ \"#{@windows_ip}:5000\" ]'"
+  root_dir = 'C:/Users/Administrator/AppData/Local/Temp'
```
sanfrancrisko, line 13: We could probably just make this more generic, seeing it's going to be used on both Windows and Unix:
```ruby
root_dir = '/foo/bar'
```
We may need to unset this in an `:after` block too, depending on the impact. However, we're currently setting `root-dir` with Windows path on Unix systems, so we may be getting away with it if `docker_args` is being set in every test (_something to look in to_)?

adrianiurca: Hi @sanfrancrisko 
The `root_dir` is set with `'C:/Users/Administrator/AppData/Local/Temp'` on windows loses and with `'root'`. But I think we can make the new test to be more generic such as:
`class { 'docker': #{docker_args}, root_dir => \"#{root_dir}/foo/bar\"}`

Kind regards,
Adrian IURCA

PR #641, opened 2020-07-28
README: add Debian 10 to list of supported OSes

kenyon:
Debian 10 is supported according to metadata.json.


daianamezdrea:
Hi @kenyon, yes, indeed Debian 10 is on the list of supported OSes, thank you! Cheers!

PR #667, opened 2020-10-13
Test against OS family rather than name

bodgit:
This then includes proper RHEL as well as CentOS. The module would (correctly) fail to work on CentOS 6 but wouldn't fail on RHEL 6.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -483,8 +483,8 @@ class docker(
     }
   }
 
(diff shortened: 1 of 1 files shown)


michaeltlombardi (approved):

PolaricEntropy:
This makes the module fail on Amazon Linux 2 (REHL7) that has os.family=RedHat but os.release.major 2. Worked fine before when it checked CentOS since then it didn't match on the fail statement.

PR #672, opened 2020-10-19
(MAINT) Use docker-compose config instead file parsing

rbelnap:
Parsing the files directly causes issues for variable interpolation in the docker-compose.yml files.  For example, the common image name `image:${IMAGE_TAG:-latest}`in a docker-compose.yml will never be found without docker-compose finding $IMAGE_TAG and substituting it.  `docker-compose config` will also merge the compose files properly without the provider needing to parse and merge manually.

My ruby is barely passable, so there may be significant improvements here.


michaeltlombardi:
Thanks so much for this contribution!

Issue #664, opened 2020-10-08
Options to docker-compose should be an Array, not a String


rbelnap:
## Use Case
Multiple options or options with parameters should be supported

```
  docker_compose { $name:
    ensure        => present,
    compose_files => [
      "${full_project_path}/${deploy_env}/docker-compose.yml",
    ],
    options => "--env-file ${full_project_path}/${deploy_env}/.env",

  }
```
Currently fails, because "--env-file path/to/env" is 2 items.  It is also not possible to specify multiple options for the same reason.  

## Describe the Solution You Would Like
This should be an array, which looks to work as expected by just changing the validation (arguments are inserted by the provider correctly already).  This would change the above to:
```
  docker_compose { $name:
    ensure        => present,
    compose_files => [
      "${full_project_path}/${deploy_env}/docker-compose.yml",
    ],
    options => ["--env-file", "${full_project_path}/${deploy_env}/.env"],
  }
```

zeha:
I'll second this, as we also need to set --env-file, and as a string this doesn't work.

zeha:
For the next person looking, this is fixed in 3.13.x and newer

PR #686, opened 2020-12-02
Options to docker-compose should be an Array, not a String

rbelnap:
This is the barest implementation for #664


adrianiurca:
HI @rbelnap, thank you for your contribution. We will close this PR, but your contribution is referenced in PR #695 because we need to update also the unit tests.

kind regards,
@adrianiurca

PR #705, opened 2021-01-12
(IAC-1218) - docker_params_changed should be run by agent

adrianiurca:


lib/puppet/functions/docker_params_changed.rb
```diff
@@ -0,0 +1,184 @@
...
+                       run_with_powershell(cmd)
+                     else
+                       cmd
+                     end
+    _stdout, _stderr, _status = Open3.capture3(create_command)
+    # logger.puts("command: #{create_command}\nstdout: #{stdout}\nstderr: #{stderr}\nstatus: #{status}\n\n")
+    # logger.puts("### END create_container\n\n")
+  end
+
+  def detect_changes(opts) # rubocop:disable Metrics/CyclomaticComplexity,Metrics/PerceivedComplexity
```
sanfrancrisko, line 90: This method is quite big - it might be good to refactor some of the code out in to separate methods or potentially create an object to handle the `docker inspect` output and allow us to query what we want from it a bit easier. E.g.
```ruby
class DockerInspectInspector
    def initialize(opts, docker_inspect_output)
      @opts = opts
      @docker_inspect_output = JSON.parse(docker_inspect_output)[0]
      @opt_volumes_count = @opts['volumes'].scan(%r{(?=:)}).count
      @os_family_windows = opts['osfamily'] == 'windows'
      @opts_volumes_valid = @opts['volumes'].is_a? String
    end

    def mounts_changed
      return unless @opts_volumes_valid
      @opts['volumes'].include?(':') && @opts['volumes'] != @docker_inspect_output['Mounts'].to_a[0]
    end

    def volumes_changed
      return unless @opts_volumes_valid
      !@opts['volumes'].include?(':') && @opts['volumes'] != @docker_inspect_output['Config']['Volumes'].to_a[0]
    end

    def volume_count_equal(i)
      return unless @opts_volumes_valid
      @opt_volumes_count == i
    end

    def param_changed?
      return true if mounts_changed && !@os_family_windows
      return true if volumes_changed && !@os_family_windows
      return true if volume_count_equal(2) && volumes_changed && !@os_family_windows
      volume_count_equal(1) && volumes_changed && @os_family_windows
    end

  end
```

...and then we can call the methods within the class to achieve what we want:
```ruby
stdout, stderr, status = Open3.capture3("docker inspect #{opts['sanitised_title']}")
docker_inspect_inspector = DockerInspectInspector.new(opts, stdout)

param_changed = docker_inspect_inspector.params_changed?
```

sanfrancrisko (commented):
Excellent work - it's been a big ticket! 😃 

From a preliminary look, it seems sound enough although I must admit it's a little bit hard to follow the `detect_changes` method as there are a lot of conditionals in here.

I suggested some potential refactor ideas to make it a bit more readable, but, just some more comments with example `docker inspect` output would help a bit more for context. It could also be useful to go over this in a Show & Tell session to step through what's happening. Unit tests may also be another way to achieve this, especially for folks not as familiar with docker and subsequently, what this method is trying to do.

lib/puppet/functions/docker_params_changed.rb
sanfrancrisko: That class is a very rough idea - do let me know if you want to go down that road before implementing. It is likely not the best way to do it, but hopefully it serves as a potential idea

lib/puppet/functions/docker_params_changed.rb
```diff
@@ -0,0 +1,184 @@
...
+  end
+
+  def remove_cidfile(cidfile, osfamily, _logger = nil)
+    delete_command = if osfamily == 'windows'
+                       run_with_powershell("del #{cidfile}")
+                     else
+                       "rm -f #{cidfile}"
+                     end
+    _stdout, _stderr, _status = Open3.capture3(delete_command)
+    # logger.puts("### BEGIN remove_cidfile\n\n")
```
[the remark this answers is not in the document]
adrianiurca: I'm planning to send the logs to the journalctl, but didn't work as I want to work(without pre-configuration) so I decide to comment them and change them sometime in the future.

lib/puppet/functions/docker_params_changed.rb
```diff
@@ -0,0 +1,184 @@
...
+  dispatch :detect_changes do
+    param 'Hash', :opts
+    return_type 'String'
+  end
+
+  def run_with_powershell(cmd)
+    "powershell.exe -Command \"& {#{cmd}}\" "
+  end
+
+  def remove_cidfile(cidfile, osfamily, _logger = nil)
```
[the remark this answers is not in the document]
adrianiurca: Also, I let the `logger` parameter to use in a 'debug' mode. If you want to see some logs from just need to uncomment the line with the specific log.

lib/puppet/functions/docker_params_changed.rb
adrianiurca: That's a great idea. I have thought to split into multiple functions, but using OOP seems to be a more readable solution.
I think this implementation will not remain in this state, it will be changed using resource API as @DavidS's recommendation.
I think all improvements will be implemented in a further ticket.
There are many improvements to be done: splitting this big function into small functions(using OOP), using ruby-pwsh,, using journalctl for logging and many other small improvements.

spec/helper/get_docker_params_changed.rb
```diff
@@ -0,0 +1,177 @@
+# frozen_string_literal: true
+
+def get_docker_params_changed(opts) # rubocop:disable Metrics/CyclomaticComplexity,Metrics/PerceivedComplexity
+  require 'open3'
+  require 'json'
+  return_value = 'NO_CHANGE_DETECTED'
```
sanfrancrisko, line 6: Would perhaps suggest
```suggestion
  return_value = 'No changes detected'
```
...to keep it consistent with the rest of the output

spec/helper/get_docker_params_changed.rb
```diff
@@ -0,0 +1,177 @@
...
+      end
+    else
+      create_container(opts['command'], opts['osfamily'], opts['image']) unless File.exist?(opts['cidfile'])
+      _stdout, _stderr, status = Open3.capture3("docker inspect #{opts['sanitised_title']}")
+      # log.puts("command: #{"docker inspect #{opts['sanitised_title']}"}\nstdout: #{stdout}\nstderr: #{stderr}\nstatus: #{status}\n\n")
+      unless status.to_s.include?('exit 0')
+        remove_cidfile(opts['cidfile'], opts['osfamily'])
+        create_container(opts['command'], opts['osfamily'], opts['image'])
+      end
+      return_value = 'NO_CHANGE_DETECTED'
```
sanfrancrisko, line 78: Would perhaps suggest:
```suggestion
      return_value = 'No changes detected'
```
...to keep it consistent with the rest of the output

spec/helper/get_docker_params_changed.rb
```diff
@@ -0,0 +1,177 @@
...
+      # log.puts(return_value)
+    end
+  else
+    return_value = 'ARG_REQUIRED_MISSING'
```
sanfrancrisko, line 82: Would perhaps suggest:
```suggestion
      return_value = 'Arg required missing'
```
...to keep it consistent with the rest of the output.

But, I'm also curious as to what we're trying to convey here?

sanfrancrisko (commented):
Just a few minor comments, given there's a ticket to address some of the other issues at a later date.

I'm still (kind of) the opinion we shouldn't commit commented out code, as good practise. If you're going to make use of the `logger` to journalctl at a later stage, I would just make it part of that commit.

sanfrancrisko (approved):

PR #725, opened 2021-02-24
Make it possible to use pod's network

seriv:
To use pod's network in a container, it shuld have --pod argument
and should not have --net.
The latter was always being added to the docker run command, if
not provided to the class as a parameter, then defaults to 'bridge'.
If empty or empty arrar - added as '' which also breaks networking
for the container in the pod.

With this fix, if a network is not defined, the whole argument will
be omitted in the docker run / docker create flags. This is the
same as providing previous default 'bridge' if container is created
or started without '--pod', and is not breaking compatibility with
podman create/run command without explicit '--net' argument.


adrianiurca (approved):
Hi @seriv, thank you for your contribution.

kind regards,
@adrianiurca

PR #738, opened 2021-03-22
Fix containers with multiple networks 

### LINKED ISSUE #679: Docker Containers with Multiple Networks Have Invalid Escaping
## Describe the Bug
The shell escaping applied at https://github.com/puppetlabs/puppetlabs-docker/blob/2f5689d77ed8dee20b05225485bd97b6270bf120/lib/puppet/parser/functions/docker_run_flags.rb#L28 is incorrect.

## Expected Behavior
Outputted script should be valid & functional. Currently script will fail with errors like: 
**Error response from daemon: network bridge --net rocketchat not found**

## Steps to Reproduce
1. Create a docker::run statement with multiple networks in array
2. Apply config

## Environment
 - Version 3.12.1
 - CentOS 7

## Additional Context
A small test case showing the incorrect escaping:

image

It seems the correct code would be something like:
    flags << "--net #{opts['net'].map{ |network| network.shellescape}.join(" --net ")}"

image

shane-davidson:
Fixes #679

### DIFF
--- a/lib/puppet/parser/functions/docker_run_flags.rb
+++ b/lib/puppet/parser/functions/docker_run_flags.rb
@@ -22,14 +22,6 @@ module Puppet::Parser::Functions
       flags << "--restart '#{opts['restart']}'"
     end
 
-    if opts['net']
-      if opts['net'].is_a? String
-        flags << "--net #{opts['net'].shellescape}"
-      elsif opts['net'].is_a? Array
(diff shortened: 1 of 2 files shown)


shane-davidson:
@pkalemba  I ended up forking the module and using this change privately because I think there was a reason this wouldn't apply to all use cases. I can't remember the exact reason though.

I think it was related to docker needing the --net command some of the time or something related to the bridge network and using the network connect command.

PR #752, opened 2021-05-31
Allow force update non-latest tagged image

yanjunding:
add an option to force update an image.

Taking the senario that, `prod` image tag is always for production deployment.
we only regularly update `prod` image tag.


adrianiurca (approved):

adrianiurca:
Hey @yanjunding, thank you for your contribution!

kind regards,
@adrianiurca

Issue #689, opened 2020-12-10
Docker 20.10 doesn't let the module work.


pier4r:
## Describe the Bug
With docker 20.10 the module (3.11.0) fails. In particular this step fails: https://github.com/puppetlabs/puppetlabs-docker/blob/v3.12.1/lib/puppet/provider/docker_network/ruby.rb#L41-L46

## Expected Behavior
That it works like with docker 19.03

## Steps to Reproduce
1. Get a centos 7.x with docker 20.10
2. have a puppet node with the docker class installed
3. let puppet run
4.  This will happen 
```
Debug: Executing: '/usr/bin/systemctl is-active docker'
Debug: Executing: '/usr/bin/systemctl is-enabled docker'
Debug: Prefetching ruby resources for docker_network
Debug: Executing: '/usr/bin/docker network ls'
Debug: Executing: '/usr/bin/docker network inspect ID'
Debug: Storing state
Debug: Stored state in 0.13 seconds
Error: Failed to apply catalog: Execution of '/usr/bin/docker network inspect ID' returned 1: WARNING: Error loading config file: .dockercfg: $HOME is not defined
[]
Error: No such network: ID
```

## Environment
 - Version 3.11.0
 - Platform Centos 7.x

## Additional Context
Docker 20.10 was installed on some of our machines yesterday 2020-12-09 and we noticed the problem popping out.

sdinten:
```
sdinten@vps04:~$ unset HOME
sdinten@vps04:/home/sdinten$ docker network ls
WARNING: Error loading config file: .dockercfg: $HOME is not defined
NETWORK ID     NAME         DRIVER    SCOPE
ec49c0686e0c   backend      bridge    local
243b554f6093   bridge       bridge    local
7a5e8f03f4eb   frontend     bridge    local
b0997cf395b1   host         host      local
2e00d2d86671   none         null      local
```
Puppet does not set $HOME;

line 41 in puppetlabs-docker/lib/puppet/provider/docker_network/ruby.rb:
lines.shift # remove header row

pelletier2017:
Also seeing this same error on Centos7 with Docker version 19.03.13, build 4484c46d9d

deubert-it:
Same problem on Debian 10 with Docker version 20.10.0, build 7287ab3

sebastianTC:
Same problem for docker volumes provider.
[...]

PR #764, opened 2021-07-28
Remove stderr empty check to avoid docker_params_changed failures when warnings appear

cedws:
Relates to #689.

This is my suggested fix for the `docker_params_changed` Deferred function in environments where `$HOME` is unset. This can happen if Puppet is ran as a service. If `$HOME` is unset, Docker prints this to stderr:

```
WARNING: Error loading config file: .dockercfg: $HOME is not defined
```

POSIX spec defines stderr as being for errors *and* diagnostic information. Therefore any output to stderr does not mean there has been an error. The check as it stands is not correct, so let's check just the status code.


adrianiurca:
Hey @cedws, thank you for your contribution. I just modified the tests accordingly with your modifications.
For me, this PR looks good!

kind regards,
@adrianiurca

adrianiurca (approved):

cedws:
Many thanks for this.

PR #770, opened 2021-08-26
Check that docker_version is not nil before using it

### LINKED ISSUE #769: `docker` fact fails when docker is not installed
## Describe the Bug

Since version 4.1.1, when the docker module is installed but Puppet runs on a machine without docker, facter fails with this error message:

```
Error: Facter: Error while resolving custom fact fact='docker', resolution='<anonymous>': undefined method `match?' for nil:NilClass
```

## Expected Behavior

The docker fact shouldn’t fail.

## Steps to Reproduce

1. Add puppetlabs/docker to your Puppetfile
2. Go to a machine without docker installed and run `puppet agent -t`
3. Puppet will show the following output:
```
$ puppet agent -t --environment production
Info: Using environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Error: Facter: Error while resolving custom fact fact='docker', resolution='<anonymous>': undefined method `match?' for nil:NilClass
Info: Caching catalog for foo.example.com
Info: Applying configuration version '1629995461'
Notice: Applied catalog in 6.52 seconds
```

## Environment

 - Version:
   * Puppet: 7.10.0
   * puppetlabs/docker: 4.1.1
 - Platform: Debian 10 (buster)

## Additional Context

This bug was introduced by 8ea5c52adac60489348b043fff3e26d4381bd9af where `match?` is used without checking that the return value from `Facter.value(:docker_client_version)` is not null (which it is on machines without docker installed).

johanfleury:
Fixes #769 and #771.

### DIFF
--- a/lib/facter/docker.rb
+++ b/lib/facter/docker.rb
@@ -68,8 +68,8 @@ end
 Facter.add(:docker_version) do
   setcode do
     if Facter::Core::Execution.which('docker')
-      value = Facter::Core::Execution.exec(
-        "#{docker_command} version --format '{{json .}}'", timeout: 90
+      value = Facter::Core::Execution.execute(
+        "#{docker_command} version --format '{{json .}}'", options: {limit: 90}
       )
       val = JSON.parse(value)
     end
@@ -80,7 +80,7 @@ end
 Facter.add(:docker_worker_join_token) do
   setcode do
     if Facter::Util::Resolution.which('docker')
-      val = Facter::Util::Resolution.exec(
+      val = Facter::Util::Resolution.execute(
         "#{docker_command} swarm join-token worker -q",
       )
     end
@@ -91,7 +91,7 @@ end
 Facter.add(:docker_manager_join_token) do
   setcode do
     if Facter::Util::Resolution.which('docker')
-      val = Facter::Util::Resolution.exec(
+      val = Facter::Util::Resolution.execute(
         "#{docker_command} swarm join-token manager -q",
       )
     end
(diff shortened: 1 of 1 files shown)


kenyon (dismissed):

johanfleury:
I added a fix for #771. I don’t know what @carabasdaniel tried to achieve in #759, but [Facter::Core::Execution.exec is deprecated](https://www.rubydoc.info/gems/facter/Facter/Core/Execution#exec-class_method) and doesn’t have a `timeout` parameter.

lib/facter/docker.rb
```diff
@@ -68,8 +68,8 @@ def interfaces
 Facter.add(:docker_version) do
   setcode do
     if Facter::Core::Execution.which('docker')
-      value = Facter::Core::Execution.exec(
-        "#{docker_command} version --format '{{json .}}'", timeout: 90
+      value = Facter::Core::Execution.execute(
+        "#{docker_command} version --format '{{json .}}'", options: {limit: 90}
```
johanfleury, line 72: Can someone who knows ruby can confirm that this is how to do it? Whatever I pass as parameter to this function is accepted.

Also, while facter’s doc says the option is called `time_limit`, [it seems to actually be called `limit` in the code](https://github.com/puppetlabs/facter/blob/187a8687cb189c0e84a8f3089f89b0ccce42523a/lib/facter/custom_facts/core/execution/base.rb#L117).

adrianiurca:
Hey @johanfleury, there seem to be some rubocop offenses. Could you please run `rubocop -a` and commit again in order to have spec tests green.

kind regards,
@adrianiurca

lib/facter/docker.rb
smortex: I guess the documentation is right and it's a bug in the code, I'll open a PR to address this :-)

There are a few more issues.  I opened #773 to address them.

johanfleury:
PR #773 supersede this one, I’ll close it.

PR #774, opened 2021-09-14
Prefer timeout to time_limit for Facter::Core::Execution

smortex:
Facter 4 documentation and code was inconsistent regarding naming of
this option.  `limit` and `time_limit` have been replaced by `timeout`
which also match the option name in previous versions of facter.  The
old names will output a warning when used, so swith to the new
terminology now.

Releated PR:
https://github.com/puppetlabs/facter/pull/2429


smortex:
Build failure unrelated.

adrianiurca (approved):

PR #792, opened 2022-02-01
Remove log_driver limitations

### LINKED ISSUE #791: Restrictions on `log_driver` prevents selection of valid log drivers
## Describe the Bug

I'm using the [loki log driver plugin](https://grafana.com/docs/loki/latest/clients/docker-driver/)  for docker, but the puppet docker module won't let me set `log_driver` to `loki`, instead it throws this error:

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Function Call, log_driver must be one of none, json-file, syslog, journald, gelf, fluentd, splunk or awslogs (file: /etc/puppetlabs/code/environments/production/modules/docker/manifests/init.pp, line: 511, column: 9) on node node.domain.tld
```

## Expected Behavior

An arbitrary string for `log_driver` would be accepted so that I could use whatever log driver I like

## Steps to Reproduce

`init.pp`

```puppet
include docker
```

`hieradata.yaml`

```yaml
docker::log_driver: loki
docker::log_opt:
  - loki-url=https://loki.domain.tld...
```

## Environment
 - Version 4.1.2
 - Platform Debain 11

timdeluxe:
Current docker versions come with a bunch of new log drivers (see https://docs.docker.com/config/containers/logging/configure/#supported-logging-drivers), most importantly "local", which is not included in the validation of this module.
With this PR i removed the validation for log_driver completely, because one can also integrate other log drivers via plugins, which should be possible.

Fixes #791

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -84,29 +84,17 @@
 # @param log_driver
 #   Set the log driver.
 #   Docker default is json-file.
-#   Valid values: none, json-file, syslog, journald, gelf, fluentd
-#   Valid values description:
-#     none     : Disables any logging for the container.
-#                docker logs won't be available with this driver.
-#     json-file: Default logging driver for Docker.
-#                Writes JSON messages to file.
-#     syslog   : Syslog logging driver for Docker.
-#                Writes log messages to syslog.
-#     journald : Journald logging driver for Docker.
-#                Writes log messages to journald.
-#     gelf     : Graylog Extended Log Format (GELF) logging driver for Docker.
-#                Writes log messages to a GELF endpoint: Graylog or Logstash.
-#     fluentd  : Fluentd logging driver for Docker.
-#                Writes log messages to fluentd (forward input).
-#     splunk   : Splunk logging driver for Docker.
-#                Writes log messages to Splunk (HTTP Event Collector).
-#     awslogs  : AWS Cloudwatch Logs logging driver for Docker.
-#                Write log messages to Cloudwatch API
+#   Please verify the value by yourself, before setting it. Valid shipped log drivers can be found here:
+#   https://docs.docker.com/config/containers/logging/configure/#supported-logging-drivers
+#   Since custom log driver plugins are and must be possible, the value can not be verified through code here.
 #
 # @param log_opt
 #   Set the log driver specific options
 #   Valid values per log driver:
 #     none     : undef
+#     local    :
+#                max-size=[0-9+][k|m|g]
+#                max-file=[0-9+]
 #     json-file:
 #                max-size=[0-9+][k|m|g]
 #                max-file=[0-9+]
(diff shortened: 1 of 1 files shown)


dploeger:
Can this be reviewed please?

LukasAud:
Closing and reopening PR to rekick automated testing.

chelnak:
@timdeluxe Thanks for this suggestion. I'm not 100% sure removing the validation is the right way to go with this.

I wonder if a better approach would be to validate against the log drivers that are currently available?

For example, the following command will return the log drivers:

```bash
docker system info --format '{{json .Plugins.Log}}'

["awslogs","fluentd","gcplogs","gelf","journald","json-file","local","logentries","splunk","syslog"]
```

One could parse the list returned and validate the user-input against it.

chelnak:
hmmm, that is an interesting analysis. I can certainly see that happening in the case where an invalid driver is passed. The question is, could we move that validation elsewhere so that it would fail on the first run...

timdeluxe:
> The question is, could we move that validation elsewhere so that it would fail on the first run...

Hit me with ideas!

How about validating it, but just issue a warning instead of a fail, if it does not fit? This way it still would be validated, but it wouldn't fail on the first run. Not perfect but would be at least a compromise...

timdeluxe:
One additional idea:
On the first run, when command you wrote above can't/doesn't return anything, because docker is not yet installed, it could validate against a hardcoded list, gathered from the current docker version and issue a warning if it doesn't fit. Then (on the next run) when the command returns something validate against that and fail, if it doesn't fit.
This way there would be no error/warning at all on a valid log_driver and a warn and then a fail on a non valid log_driver.

chelnak:
I think that would actually be a good non destructive option. However it's incorrect, we should warn and fall back to a known default..

chelnak:
I think that would actually be a good non destructive option. However it's incorrect, we should warn and fall back to a known default..

chelnak:
> Then (on the next run) when the command returns something validate against that and fail, if it doesn't fit.

Yeah this is certainly another way to approach it. However I'm not a massive fan of things taking more than one puppet run to apply unless we have no choice.

Warning and falling back to a known default seems more reasonable here to me.

What do you think?

chelnak:
Hey @timdeluxe , did you get chance to think over my suggestion?

timdeluxe:
@chelnak sorry for missing feedback, was busy with some actions around my own (round) birthday, please apologize.

I am sorry, but i don't get you at...

> Warning and falling back to a known default seems more reasonable here to me.

...what do you mean with "falling back to a known default"? Do you mean the validation should just been done to a known set of log_drivers like done in the current code but the set just will extend with f.e. the currently missing "local" one?
Or do you mean that the log_driver setting itself should be changed by the code if it does not validate against a hardcoded list?
Well, i also don't like the two puppet runs issue, but we have a really crazy case here. When thinking about "known defaults" please always remember, that one may want to use own logging drivers via (self developed maybe) plugins - no "known default" would be ever able to cover this and if a warning would be issued in this case, one has to live with a warning forever - i also do not like that.

In my opinion the only alternatives are:
a) remove the validation at all, like done originally in this PR
b) do as i proposed with warning against a known list and a fail, if the command for the validation gives needful results on later puppet runs

But maybe i did not understand you at some point - in that case, please explain in more detail. Thanks!

timdeluxe:
@chelnak Any update here?

chelnak (approved):

chelnak:
@timdeluxe Thanks for getting this updated.

The test failures are not related to this PR so I am happy to get this merged.

chelnak (approved):
LGTM - thank you!

PR #799, opened 2022-02-15
(IAC-1729) Add Support for Debian 11

david22swan:


david22swan:
Created custom workflow to test Debian 11, the changes worked as shown here:
https://github.com/puppetlabs/puppetlabs-docker/actions/runs/188301700
Screen Shot 2022-02-22 at 5 43 11 PM
Whether the test framework should be expanded to included them by default, or if a custom job should be created to run all tests, but only when manually kicked, is for another time

chelnak (approved):

PR #805, opened 2022-03-23
Fix idempotency when using scaling with docker-compose

canihavethisone:


canihavethisone (commented):
fixed lint

canihavethisone:
Can I get tests run on this again please?  I have fixed the lint, but am also looking at the acceptance test as it doesn't include scaling for docker-compose.  Will look at that unless someone more familiar with this module can do it faster.

canihavethisone:
There appear to be the error below on some tests.  Any insight on if this needs to be stubbed or managed otherwise?

```
#   Permission denied @ dir_s_mkdir - /tmp_docker
#   ./lib/puppet/provider/docker_compose/ruby.rb:17:in `mkdir'
```

canihavethisone:
I have tried using the fileutils resource however the above issue still occurs, only in testing.  My guess is that the test is trying to create the /tmp_docker folder on the testing host, and failing.  Any advice appreciated

chelnak:
@canihavethisone Hey thanks! I think i'm with you regarding the `root` directory.. maybe we should aim for a safer directory.

@david22swan Have you seen anything like this before?

canihavethisone:
@chelnak  thanks for assisting with this.  I have spend ages trying to mock/stub rspec to not create the /tmp_docker folder during spec test, I think that whatever folder is used we don't want rspec shelling out to actually create it!  I can't get a stub to work, but feel it is the way to get it passing.

canihavethisone:
Now added a successful stup for /tmp_docker.  Can you please re-run tests, and merge if acceptable.  Thanks.

chelnak:
@canihavethisone Thanks for this. Tests are running!

canihavethisone:
@chelnak : Interesting, it looks like the spec test for define docker::registry is failing now.  While the stub could be added to that test, I am unsure why it is referencing the mkdir from the compose?

chelnak:
I'm not sure off the top of my head. I'd have to pull the code down and take a look. I'll try and get to that today but I'll also ping this to the team. Someone might be able to spot something.

canihavethisone:
Thanks.  I'm running spec tests with 'pdk test unit' and that wasn't failing locally, perhaps it's only running tests in /spec and not /functions

canihavethisone:
@chelnak thanks for the parallel_spec command.  I have got all spec tests to pass now by:

1. moving the stub from the spec to the unit/lib/puppet/type test
2. adding another stub to the init spec and added
3. adding the stub to the defines/registry_spec.rb   <--- still unsure why the registry define needed the dir stub!

Look like it should be fine now, although I did see one of the acceptance tests fail on connection error previously which seems unrelated.  Boy, I don't like the way rspec stub work!

chelnak:
Great job! Thanks for putting in the work!!

canihavethisone:
No worries.  I'm trying to fix 2 issues in this merge request:

1. the known 'docker_compose can't run execs in /tmp unless remounted' issue
2. scaling in docker compose wasn't idempotent, or tested in acceptance

I appreciate the opportunity to get this working.  A colleague actually helped me with the stub syntax as it was stuck.

canihavethisone:
@chelnak  Passed! :)

chelnak (approved):

chelnak:
Thanks for your contribution! 🚀

lib/puppet/provider/docker_compose/ruby.rb
```diff
@@ -13,6 +13,11 @@
     environment(HOME: '/root')
   end
 
+  has_command(:docker_compose, command(:dockercompose)) do
+    Dir.mkdir('/tmp_docker') unless Dir.exist?('/tmp_docker')
```
kenyon, line 17: I'm getting permission denied from this, even when puppet agent is running as root. Besides that, it seems wrong to be creating this directory in the root of the filesystem. Also, what is this for? The `TMPDIR` env var is never used in this code. Seems like this is here by mistake for the tests only.

canihavethisone: 1. Unsure why you are getting permission denied, it works for me IRL and also passes acceptance tests.  What OS are you using?
2. /tmp_docker is in the base path so that it is adjacent to /tmp, which is the dir that docker-compose uses by default.  What location do you suggest?
3. the $TMPDIR env var is respected by the Python underlying docker-compose.  Using a path other than /tmp and setting that var overcomes the well-documented issue of docker-compose failing to run when noexec is set on /tmp.  The change is there to overcome this and not for testing purposes only.

kenyon: On Ubuntu. What is the issue that this pull request is trying to address?

spec/classes/init_spec.rb
```diff
@@ -196,6 +196,12 @@
               }
             else
               it {
+                # Stub /tmp_docker dir to prevent shelling out during spec test
```
kenyon, line 199: My guess is that these are hiding real errors.

lib/puppet/provider/docker_compose/ruby.rb
kenyon: I wonder if the problem is that this code is running on the Puppet Server, not the agent, because the error occurs during catalog compilation. I can't find any evidence of this error other than during the puppet agent run, so I'm not sure why it's getting permission denied.

kenyon: Also, it would be good if this were somehow made optional, so that we don't end up with `/tmp_docker` on every system, even where docker isn't used.

spec/classes/init_spec.rb
canihavethisone: This stub/mock is to address the spec test only, not acceptance tests (which passed)

lib/puppet/provider/docker_compose/ruby.rb
canihavethisone: @kenyon to address your comments:
- the pull request addresses the known issue with docker-compose executing from /tmp as also reported at https://github.com/docker/compose/issues/8041
- agree that the issue, while appearing only on the first puppet run, may be trying to apply the dir on the server.  It may be linked to other reported issues as referenced on your issue raised as a result of this unexpected outcome
- the actual creation of the dir only occurs if resources are applied by docker_compose, and will not actually occur otherwise.  Further, if you have this module on your puppetserver but not applied to nodes, they do not appear to be affected.

I agree that it is both unexpected and need to be addressed, and have reported more detailed observations in your raised issue and also requested maintainer input.

Thank you for reporting the issue.

PR #807, opened 2022-03-23
pdksync - (GH-iac-334) Remove Support for Ubuntu 14.04/16.04

david22swan:
(GH-iac-334) Remove Support for Ubuntu 16.04
pdk version: `2.1.0` 

- Fix put in place for the fixtures.yml
- Removed support for Ubuntu 14.04, seem's to have been missed during an earlier removal sweep

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -5,9 +5,9 @@ fixtures:
     powershell: 'puppetlabs-powershell'
     reboot: 'puppetlabs-reboot'
   repositories:
-    facts: 'git://github.com/puppetlabs/puppetlabs-facts.git'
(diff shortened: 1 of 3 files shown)


LukasAud (approved):

bastelfreak (changes requested):
https://github.com/puppetlabs/puppetlabs-docker/search?q=14.04 also here the module contains some 14.04/16.04 specific code and I really think that should be cleaned up in combination with the metadata.json remove of those operating systems. Otherwise it's really complicated in the future to determine if a change is a breaking change or not.

david22swan:
@bastelfreak This is just a quick automated pr to drop support, will be going through each module afterwards to remove compatibility.

LukasAud (approved):

PR #811, opened 2022-03-28
WIP: docker_run_flags: Shellescape any provided values

neomilium:


chelnak:
Hey 👋 

Can you set this PR to ready for review?

thanks!

smortex:
I am wondering if the CI failures are relevant… Like `volume-1:C:\volume_1` being escaped as `volume-1:C:\\volume_1`…  Not sure about how to check this…

neomilium:
@chelnak Test added by @smortex, PR is now ready.

chelnak:
@smortex - i'll take a closer look at this tomorrow. Seems odd that all of the failures here are for Windows and they don't seem to be on the nightly tests either..

smortex:
@neomilium @chelnak I added a commit that seems to help… but escaping on windows is total nonsense and I am sure of nothing :scream: 

Some windows node pass, the one failing seems to fail earlier than the code added / changed here and the failures might be the result of some external service not being reliable… maybe?

lib/puppet/parser/functions/docker_run_flags.rb
```diff
@@ -5,17 +5,28 @@
 # docker_run_flags.rb
 #
 module Puppet::Parser::Functions
+  newfunction(:'docker::escape', type: :rvalue) do |args|
+    subject = args[0]
+
+    #scope = closure_scope
```
smortex, line 11: Looks like I forgot to remove this comment :clown_face:.  Don't forget to drop it when this is ready for merging.

lib/puppet/parser/functions/docker_run_flags.rb
```diff
@@ -5,17 +5,28 @@
...
+    if self['facts'] && self['facts']['kernel'] == 'windows'
```
smortex, line 12: The rest of the code seems to much more check for `os[:family] == 'windows'`

lib/puppet/parser/functions/docker_run_flags.rb
```diff
@@ -5,17 +5,27 @@
 # docker_run_flags.rb
 #
 module Puppet::Parser::Functions
+  newfunction(:'docker::escape', type: :rvalue) do |args|
+    subject = args[0]
+
+    if self['facts'] && self['facts']['os']['family'] == 'windows'
+      %("#{subject.gsub('"', '\\"')}")
```
bastelfreak, line 12: is that just because the shellescape function is designed for linux? what about patching https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/parser/functions/shell_escape.rb instead?

smortex: The more I am thinking about it, the more I am concerned about cmd.exe escaping vs. powershell escaping: they seem to be different in incompatible ways:

* https://ss64.com/nt/syntax-esc.html
* https://ss64.com/ps/syntax-esc.html

Maybe the stdlib should have 3 shell escape functions: `shell_escape`, `powershell_escape` and `cmd_escape`, then modules could have a local escape function which call the right escape function on windows depending on what the module is using?

smortex, line 12: The escaping on line 68 suggest PowerShell escaping is used (backtick at end of line while cmd.exe would use a `^`):

https://github.com/puppetlabs/puppetlabs-docker/blob/355b5b33714cd82f91e24bf400aefde2515b1a74/lib/puppet/parser/functions/docker_run_flags.rb#L68

so quotes escaping above should be wrong… but I can't give it a try on windows ATM

bastelfreak: that sounds like a good idea

smortex: https://github.com/puppetlabs/puppetlabs-stdlib/pull/1235 was merged in stdlib. When we have a new release with this code, we will be able to move on with this PR.

smortex: Awesome! I added a commit to rely on this new code.  Let's see what the CI think about it :wink:

lib/puppet/parser/functions/docker_run_flags.rb
```diff
@@ -5,17 +5,27 @@
 # docker_run_flags.rb
 #
 module Puppet::Parser::Functions
+  newfunction(:'docker::escape', type: :rvalue) do |args|
+    subject = args[0]
+
+    if self['facts'] && self['facts']['os']['family'] == 'windows'
+      call_function('powershell_escape', subject)
+    else
+      subject.shellescape
```
smortex, line 14: For this to be more readable, maybe remove `require 'shellwords'` above and change this to:

```suggestion
      call_function('shellescape', subject)
```

smortex:
Last CI failures seems to be caused by inaccessible windows test nodes. I think this is ready for review.

chelnak:
Failed jobs have been restarted. If everything looks ok I’d like to get this in to the next release.

bastelfreak (approved):

canihavethisone:
@smortex @chelnak why has stdlib been raised to min v8.2.0 in the pull request?  That is very high and will have breaking impact when combined with a number of other modules that have a max lower or need to maintain compatibility with lower versions of stdlib.

smortex:
> @smortex @chelnak @bastelfreak - why has stdlib been raised to min v8.2.0 in the pull request? That is very high and will have breaking impact when combined with a number of other available modules

Because we had to add escape functions for windows in the stdlib: see discussion above and https://github.com/puppetlabs/puppetlabs-stdlib/pull/1235.

chelnak:
@smortex @bastelfreak How do you both feel about moving forward with this PR? 

Pending test failures which I think may be transient, I think it is good.

Though I do wonder if it is a breaking change due to the higher requirements of `stdlib`.

What are your thoughts?

chelnak (approved):

canihavethisone:
My concern is that the major fast forward of the stdlib dependency may have significant impact for users if this module in deployments alongside other modules that haven't caught up yet. I recon a good metric would be how many popular modules still have stdlib pinned lower than v.8x, and weigh that against the urgency of this change or if it can be merged when 8x is more commonly scoped.  I understand there will be diversity of opinions but think its worth considering.

neomilium:
> I recon a good metric would be how many popular modules still have stdlib pinned lower than v.8x, and weigh that against the urgency of this change or if it can be merged when 8x is more commonly scoped.

How to define "popular modules"?

smortex (approved):
> How do you both feel about moving forward with this PR?

My position is that we have tooling to check module versions and that roles and profile in a control-repo are working as expected (e.g. using onceover), so do not worry much about bumping versions. I would not expect users to blindly update things, skipping major versions and assume everything will be fine witout testing :trollface:.

canihavethisone:
Hi, im not concerned about people blindly updating, I'm concerned that we will lock out users who use this module alongside numerous others. Locking them out will deny them of new features and bugfixes until their other dependencies support stdlib v8.  Bottom line IMO is that there is a huge difference between setting the upper limit to v8 or the lower *requirement* to v8.  Its also a cost v benefit consideration at this moment in v8's youth.

binford2k:
Here's a list of all the Forge modules that pin their stdlib dependencies to <8.x, sorted by download count (a weak correlation to popularity.) Notice that many of them are deprecated.

https://docs.google.com/spreadsheets/d/16QT1wd5CmwJL9TbKe4GZgpPi3W4nXHo7Q4fD5aK8bRM/edit#gid=223629225

LukasAud:
Hi @neomilium. After careful evaluation of your PR, I'm happy to inform you that we are ready to go ahead and merge it. Thanks for your contribution!

We also noted that there are some concerns about this PR, such as the one raised by @canihavethisone. We have considered the potential impact of these changes and we have decided to still go forward with it. While it is true that there may be some compatibility issues with a few non-supported modules, we also have noted that this change may be important for one of our current security enhancement projects and, as such, either this or something very similar was bound to be implemented sooner or later. We are sorry if this causes any inconvenience.

LukasAud:
Hey @neomilium. We are sorry for this inconvenience but we had to revert your PR in order to follow best practices when releasing a new version of our module, by making a minor release prior to a backwards incompatible one. We could re-implement it but I'm afraid it would appear under our name. Could you reopen your branch under a new PR so that we can merge and give proper credit to all contributors?

kenyon:
Rebasing or cherry-picking the commits would maintain the original commit authors.

PR #812, opened 2022-03-30
Support different Architectures (like `aarch64`) when installing Compose

mpdude:
Note that `aarch64` is only supported for Compose V2.


smortex (approved):
LGTM

PR #820, opened 2022-04-12
Fix permission denied issue introduced in v4.2.0

### LINKED ISSUE #819: Permission denied from mkdir /tmp_docker in v4.2.0
With v4.2.0 of this module, I get `Permission denied - /tmp_docker` from a puppet agent run.

Caused by #805: https://github.com/puppetlabs/puppetlabs-docker/pull/805/files#r847521695

chelnak:
# Context

This PR fixes #819 

# What has changed?

This PR reverts the modification of the TMPDIR environment variable from the previous release. This change was made to fix a bug in docker compose where some processes would fail if the noexec bit had been set on /tmp. However this change caused unexpected failures in certain environments.

### DIFF
--- a/lib/puppet/provider/docker_compose/ruby.rb
+++ b/lib/puppet/provider/docker_compose/ruby.rb
@@ -14,8 +14,7 @@ Puppet::Type.type(:docker_compose).provide(:ruby) do
   end
 
   has_command(:docker_compose, command(:dockercompose)) do
-    Dir.mkdir('/tmp_docker') unless Dir.exist?('/tmp_docker')
-    ENV.store('TMPDIR', '/tmp_docker')
+    environment(HOME: '/root')
   end
 
   def exists?
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -196,12 +196,6 @@ describe 'docker', type: :class do
               }
             else
               it {
-                # Stub /tmp_docker dir to prevent shelling out during spec test
-                allow(Dir).to receive(:exist?).and_wrap_original do |original_method, a|
-                  original_method.call(a)
-                end
-                allow(Dir).to receive(:exist?).with('/tmp_docker').and_return(true)
-
                 is_expected.to contain_class('docker::repos').that_comes_before('Class[docker::install]')
                 is_expected.to contain_class('docker::install').that_comes_before('Class[docker::config]')
                 is_expected.to contain_class('docker::config').that_comes_before('Class[docker::service]')
--- a/spec/defines/registry_spec.rb
+++ b/spec/defines/registry_spec.rb
@@ -2,15 +2,6 @@
 
 require 'spec_helper'
 
-# Stub /tmp_docker dir to prevent shelling out during spec test
-class Dir
-  class << self
-    def exist?(var)
-      return true if var == '/tmp_docker'
-    end
-  end
-end
-
 tests = {
   'with ensure => absent' => {
     'ensure' => 'absent',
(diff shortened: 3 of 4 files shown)


lib/puppet/provider/docker_compose/ruby.rb
```diff
@@ -14,8 +14,7 @@
   end
 
   has_command(:docker_compose, command(:dockercompose)) do
-    Dir.mkdir('/tmp_docker') unless Dir.exist?('/tmp_docker')
-    ENV.store('TMPDIR', '/tmp_docker')
+    environment(HOME: '/root')
```
smortex, line 17: Overriding $HOME unconditionally is probably a bad idea. This is still a draft so probably needed for some tests, but flag it so that we don't forget about it by mistake.

chelnak: This seems to be true for other command definitions. It's a temporary override but I do see your point.

bjvrielink: Please please please do not create new directories in /.
Please pick some other place (for example under /usr/local/share, /opt, /var/cache, or anywhere else but /).

chelnak: Hey @bjvrielink, I think you may have misinterpreted the proposed change here. 

This PR removes the directory creation.

bjvrielink: Yeah, now I noticed. I've spent a little too much time right now trying to figure out what was/is broken in the current release.

chelnak: We are working on a fix 🙂

For now, I would pin to the previous version.

canihavethisone: I did some runtime tests with the new changes, and line 27 of lib/puppet/provider/docker_compose/ruby.rb was failing as it wasn;t getting the new tmpdir param.

I got it to work by changing line 27 of lib/puppet/provider/docker_compose/ruby.rb to
```
compose_output = YAML.safe_load(exec_dockercompose(args))
```
however, it is not idempotent as the TMPDIR var gets set every time the resource runs.  I can't find a way to silence it.

chelnak: Good spot! 

Interesting regarding the idempotency. I wouldn't have expected it to have an impact.

It's also highly possible that my suggested implementation is not the right fit here.

Maybe we shouldn't be overridden TMPDIR at all given that it it isn't actually a docker setting.

Could this boil down to a simple documentation update where we advise that impacted users may want to override TMPDIR with another puppet resource?

@smortex @kenyon what are your thoughts?

canihavethisone: I tried setting the TMPDIR env var previously using another puppet resource.  However without a reboot the docker_compose resource wasn't seeing/respecting it.  I like your draft approach as it takes care of it close to the compose resource, it just needs to occur silently if the param is declared - interestingly the existing (breaking) approach did that, however my testing with a hybrid method wasn;t reading the resource[:tmpdir] if set earlier.

canihavethisone: An alternative approach may be to precede the docker-compose command with 'TMPDIR=resource[:tmpdir]' so that it occurs as part of the command if the tmpdir param is declared.  That way ruby isn't doing it, the shell is.  Just a thought.

david22swan (approved):

PR #823, opened 2022-04-15
Add tmpdir option to docker_compose

canihavethisone:
This pull request adds an optional tmpdir parameter to the docker_compose resource.  It is useful on systems where the default /tmp directory is mounted with noexec, or consumers wish to specify an alternate tmpdir other than the system-wide environment variable.  Note that it does **_not_** create the target directory, the consumer is responsible to ensure that the directory exists and is executable.

As well as adding the type from @chelnak  example, I have added the following to the provider, which includes verification that the tmpdir target exists, warns and does not set the env var if not.  If there is a better way to perform this in Ruby, please comment.

```Ruby
  def set_tmpdir
    return unless resource[:tmpdir]
    # Check if the the tmpdir target exists
    Puppet.warning("#{resource[:tmpdir]} (defined as docker_compose tmpdir) does not exist") unless Dir.exist?(resource[:tmpdir])
    # Set TMPDIR environment variable only if defined among resources and exists
    ENV['TMPDIR'] = resource[:tmpdir] if Dir.exist?(resource[:tmpdir])
  end
```

The changes pass unit tests and have been verified with the following manifest
```puppet
class test {

  file { '/usr/local/share/tmp_docker':
    ensure => directory
  }

  class { 'docker':
    log_driver => 'journald',
  }
  
  $compose = "compose_test:
    image: ubuntu:14.04
    command: /bin/sh -c 'while true; do echo hello world; sleep 1; done'
  "
  
  file { '/tmp/compose.yml':
    content => $compose
  }
  
  class { 'docker::compose':
    ensure => present,
  }
  
  docker_compose {'test':
    ensure => present,
    compose_files => [
      '/tmp/compose.yml'
    ],
    tmpdir => '/usr/local/share/tmp_docker',
    scale  => {
      compose_test => 3
    },
  }
  docker_compose {'test2':
    ensure => present,
    compose_files => [
      '/tmp/compose.yml'
    ],
  }
}
```
**Note1**:  this implementation is optional, and does not affect global TMPDIR environment variable (as tested on Linux) as it only applies within the docker_compose resource.  I have also updated the readme to reflect that this is optional.

**Note2**: if any resource has the tmpdir param set, subsequent resources will inherit it unless they specify a different tmpdir target.  Therefore it only has to be specified in the first resource if you want it to apply to all (as per above example, test2 will also use the same tmpdir, however if test2 was declared first it would use the system TMPDIR variable or default to /tmp if not set and fail if that dir is noexec).  If subsequent resource tmpdir is defined but does not exist, a warning is raised and the previously successful tmpdir value is used.

Apologies for the previous attempt at this that also managed directory creation and introduced a breaking change on first puppet run (though other changes in that merge did fix idempotency with scaling).  This however is a more considered approach to address issue as reported at https://github.com/docker/compose/issues/8041 and experienced on hardened systems with noexec on /tmp.  

I have not expanded the acceptance test for these changes, and defer to maintainers if they wish to add one.

Any comment or suggestions are welcome.

### DIFF
--- a/lib/puppet/provider/docker_compose/ruby.rb
+++ b/lib/puppet/provider/docker_compose/ruby.rb
@@ -22,7 +22,7 @@ Puppet::Type.type(:docker_compose).provide(:ruby) do
       # Check if the the tmpdir target exists
       Puppet.warning("#{resource[:tmpdir]} (defined as docker_compose tmpdir) does not exist") unless Dir.exist?(resource[:tmpdir])
       # Set TMPDIR environment variable only if defined among resources and exists
-      ENV['TMPDIR'] = resource[:tmpdir] unless !resource[:tmpdir] unless !Dir.exist?(resource[:tmpdir])
+      ENV['TMPDIR'] = resource[:tmpdir] unless !Dir.exist?(resource[:tmpdir])
     end
   end
 
--- a/lib/puppet/provider/docker_compose/ruby.rb
+++ b/lib/puppet/provider/docker_compose/ruby.rb
@@ -17,10 +17,17 @@ Puppet::Type.type(:docker_compose).provide(:ruby) do
     environment(HOME: '/root')
   end
 
+  def set_tmpdir
+    # Set TMPDIR environment variable only if defined among resources
+    ENV['TMPDIR'] = resource[:tmpdir] unless !resource[:tmpdir]
+  end
+
   def exists?
     Puppet.info("Checking for compose project #{name}")
     compose_services = {}
     compose_containers = []
+
+    set_tmpdir
 
     # get merged config using docker-compose config
     args = [compose_files, '-p', name, 'config'].insert(3, resource[:options]).compact
--- a/lib/puppet/provider/docker_compose/ruby.rb
+++ b/lib/puppet/provider/docker_compose/ruby.rb
@@ -18,8 +18,12 @@ Puppet::Type.type(:docker_compose).provide(:ruby) do
   end
 
   def set_tmpdir
-    # Set TMPDIR environment variable only if defined among resources
-    ENV['TMPDIR'] = resource[:tmpdir] unless !resource[:tmpdir]
+    if resource[:tmpdir]
+      # Check if the the tmpdir target exists
+      Puppet.warning("#{resource[:tmpdir]} (defined as docker_compose tmpdir) does not exist") unless Dir.exist?(resource[:tmpdir])
+      # Set TMPDIR environment variable only if defined among resources
+      ENV['TMPDIR'] = resource[:tmpdir] unless !resource[:tmpdir] unless !Dir.exist?(resource[:tmpdir])
+    end
   end
 
   def exists?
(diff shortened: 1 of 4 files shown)


canihavethisone:
Further comment, I found that the `pdk validate` command presented a large number of warnings and errors in the module outside of these changes.  I have not addressed these but bring them to the attention of module maintainers for their consideration.

canihavethisone:
I'm thinking it would also be possible to add evaluation and warn if the following fail:
1. tmpdir is not set in Puppet & tmpdir is not set in system & /tmp is noexec on system
2. tmpdir is set in Puppet but is noexec
3. tmpdir is set in System but is noexec

However, this may be going a bridge to far.

chelnak:
Hey! It's taken ages to get round to this sorry!

I think that directory management should be left to the user.

We already have the tools available for this at makes sense to keep the domains isolated.

I'd be interested in getting some other opinions on this change too. 

Personally I think it's a reasonable to have this as an optional part of the configuration but there may be other suggestions out there.

Thanks for putting in the effort 😀

canihavethisone:
@chelnak yep, the pull request as it stands does not attempt directory creation and only warns if the (option) declared tmpdir does not exist.  It achieves the desired outcome without the issue from the previous merge.  I'll delete my subsequent comment with the example of creating the dir to avoid confusion.

lib/puppet/provider/docker_compose/ruby.rb
```diff
@@ -17,11 +17,21 @@
     environment(HOME: '/root')
   end
 
+  def set_tmpdir
+    return unless resource[:tmpdir]
+    # Check if the the tmpdir target exists
+    Puppet.warning("#{resource[:tmpdir]} (defined as docker_compose tmpdir) does not exist") unless Dir.exist?(resource[:tmpdir])
+    # Set TMPDIR environment variable only if defined among resources and exists
+    ENV['TMPDIR'] = resource[:tmpdir] if Dir.exist?(resource[:tmpdir])
```
smortex, line 25: How/when is this variable reset to it's original value?

I suspect that this would have an impact with other types / provider depending on resource ordering:

```puppet
foo { 'before': # ENV['TMPDIR'] #=> nil
}

docker_compose { 'bar':
  path => '/here/or/there',
}

foo { 'after': # ENV['TMPDIR'] #=>  '/here/or/there'
}
```

canihavethisone: @smortex the variable only appears to have effect within the resource, and also the resource collector in init.pp ensures that docker_compose runs lasts, so my tests to put (echo) the ENV['TMPDIR'] during run show it as unset or reflect the system variable within prior resources
```
    Class['docker::repos']
    -> Class['docker::install']
    -> Class['docker::config']
    -> Class['docker::service']
    -> Docker::Registry <||>
    -> Docker::Image <||>
    -> Docker::Run <||>
    -> Docker_compose <||>
```

chelnak: @canihavethisone Thanks for that explanation. Makes sense to me.

@smortex What do you think?

chelnak (approved):

chelnak:
Merging - changes are now optional and should not impact existing users.

PR #825, opened 2022-04-25
Support Ubuntu 22.04

mpdude:
This declares Ubuntu 22.04 as compatible in the `metadata.json` file.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -36,7 +36,8 @@
       "operatingsystem": "Ubuntu",
       "operatingsystemrelease": [
         "18.04",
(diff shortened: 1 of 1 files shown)


david22swan:
@mpdude Hey, I'm sorry but we can't support Ubuntu 22.04 at this time so I'm gonna have to close this.
This is due to the fact that it is not yet supported by the puppet agent and thus we can't effectively test against it, preventing us from claiming support.
Once this has been resolved however we will be sure to update support on all of our modules as quickly as possible.

PR #833, opened 2022-05-23
Fix docker-compose, network and volumes not applying on 1st run, fix other idempotency

canihavethisone:
These changes appear to fix the issue where the docker and docker-compose binaries are not found on first run, by removing the early realizing of those resources before they actually exist.

A further issue with selinux context being reset on container `.system` files,  `service-overrides.conf` and `socket-overrides.conf` is also fixed by setting the seltype upon creating these files.  This also prevents containers being restarted as a result.

Despite the removal of `commands` in the provider files, the `has_command` appears to be sufficient to allow the binaries to be located at runtime and after first installation.  The `HOME: '/root'` environment parameter has also been removed from the providers without perceived side-effect.

I have tested these changes to the desired effect, however welcome others to test and verify please.

### DIFF
--- a/lib/puppet/provider/docker_compose/ruby.rb
+++ b/lib/puppet/provider/docker_compose/ruby.rb
@@ -6,16 +6,10 @@ Puppet::Type.type(:docker_compose).provide(:ruby) do
   desc 'Support for Puppet running Docker Compose'
 
   mk_resource_methods
-  commands dockercompose: 'docker-compose'
-  commands dockercmd: 'docker'
 
-  has_command(:docker, command(:dockercmd)) do
-    environment(HOME: '/root')
-  end
+  has_command(:docker, 'docker')
 
-  has_command(:dockercompose, command(:dockercompose)) do
-    environment(HOME: '/root')
-  end
+  has_command(:dockercompose, 'docker-compose')
 
   def set_tmpdir
     return unless resource[:tmpdir]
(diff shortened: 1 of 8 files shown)


canihavethisone:
@chelnak @shoddyguard @marrajo-afk @jhaezebr has anyone been able to test and verify the expected fix with this pull request?  It looks good to me though I would like some peer review please.

chelnak:
Hey! It's community day today so I'll add it to my list.

It would speed up my repro if you could provide a really basic manifest!

canihavethisone:
I think the same test manifest from my last pull request should do.  It will install the newer default version of docker-compose, and with the changes in this request will apply on first puppet run instead of requiring 2.  The manifest also verifies that my last 2 merged features are unaffected - scaling fixed and use of optional tmpdir param

```
class test {

  file { '/usr/local/share/tmp_docker':
    ensure => directory
  }

  class { 'docker':
    log_driver => 'journald',
  }
  
  $compose = "compose_test:
    image: ubuntu:14.04
    command: /bin/sh -c 'while true; do echo hello world; sleep 1; done'
  "
  
  file { '/tmp/compose.yml':
    content => $compose
  }
  
  class { 'docker::compose':
    ensure => present,
  }
  
  docker_compose {'test':
    ensure => present,
    compose_files => [
      '/tmp/compose.yml'
    ],
    tmpdir => '/usr/local/share/tmp_docker',
    scale  => {
      compose_test => 3
    },
  }
  docker_compose {'test2':
    ensure => present,
    compose_files => [
      '/tmp/compose.yml'
    ],
  }
}
```

shoddyguard:
Pulled this module into my Dev environment and it worked perfectly (Ubuntu 18.04 & 20.04).

chelnak:
LGTM! 🥳 

image

chelnak (approved):

PR #841, opened 2022-06-07
New create_user parameter on main class

### LINKED ISSUE #840: docker::system_user::create_user is not exposed.
The `create_user` parameter of the `docker::system_user`is not really usable. In particular as it it is currently called from the `docker::config class`.

I expect a new parameter in the main class is needed that can be passed through.

traylenator:
Currently when specifying `docker_users` every user is
both created and added to the docker group.

If `create_user` is false then the users are presumed to exist
via another method outside this module and will only be added to
the docker group.

Fixes #840

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -1,9 +1,12 @@
-# @summary
+# @summary Configuration for docker
+# @api private
 #
 class docker::config {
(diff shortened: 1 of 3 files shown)


traylenator:
CLA has been signed on behalf of CERN IT with PuppetLabs.

Issue #842, opened 2022-06-13
Rebuild config.json, if not present


maltewhiite:
If I delete `/root/.docker/config.json` then it isn't rebuilt by docker::registry. 

~~I suggest adding a `    creates     => '/root/.docker/config.json',` to the exec.~~

https://github.com/puppetlabs/puppetlabs-docker/blob/f7df65a9c2eff754e798ff6c21f22e260662ffc8/manifests/registry.pp#L150

I am using version 4.2.1 of the module. 

<details><summary>Removing file, then running puppet.</summary>
 
```shell
[:~/.docker]# ll
total 4
-rw-------. 1 root root 413 Jun 13 12:34 config.json
[:~/.docker]# mv config.json config.json_backup
[:~/.docker]# ll
total 4
-rw-------. 1 root root 413 Jun 13 12:34 config.json_backup
[:~/.docker]# puppet agent -t
Info: Using environment 'test'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Retrieving locales
Info: Loading facts
Info: Caching catalog for 
Info: Applying configuration version '1655125098'
Info: Checking if docker volume gitlab-runner-config exists
Notice: Applied catalog in 9.65 seconds
[:~/.docker]# ll
total 4
-rw-------. 1 root root 413 Jun 13 12:34 config.json_backup
[:~/.docker]# 
```

</details>

This is the puppet code we use to set up connection to our private docker registry. 

```ruby
  # Use our private Docker Registry
  docker::registry { $docker_registry:
    server   => $docker_registry,
    username => $repository_username,
    password => $repository_password,
    require  => Class['docker'],
  }
```
<details><summary>code comments for "creates"</summary>

<h2>creates</h2>
<p>A file to look for before running the command. The command will
only run if the file <strong>doesn’t exist.</strong></p>

<p>This parameter doesn’t cause Puppet to create a file; it is only
useful if <strong>the command itself</strong> creates a file.</p>

<pre><code>exec { 'tar -xf /Volumes/nfs02/important.tar':
   cwd     =&gt; '/var/tmp',
   creates =&gt; '/var/tmp/myfile',
   path    =&gt; ['/usr/bin', '/usr/sbin',],
 }
</code></pre>
 
<p>In this example, <code>myfile</code> is assumed to be a file inside
<code>important.tar</code>. If it is ever deleted, the exec will bring it
[...]

PR #843, opened 2022-06-13
Simplify idempotency check

maltewhiite:
If the receipt is only for idempotency, then we can simply use "creates" instead. 

I am not sure if I am fulfilling all requirements here. 

Closes https://github.com/puppetlabs/puppetlabs-docker/issues/842

### DIFF
--- a/manifests/registry.pp
+++ b/manifests/registry.pp
@@ -32,15 +32,15 @@
 # @param version
 #
 define docker::registry (
-  Optional[String]               $server          = $title,
-  Optional[Enum[present, absent]] $ensure         = 'present',
-  Optional[String]               $username        = undef,
-  Optional[String]               $password        = undef,
-  Optional[String]               $pass_hash       = undef,
-  Optional[String]               $email           = undef,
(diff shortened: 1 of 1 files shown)


maltewhiite:
Last commit was to fix failing check: https://github.com/puppetlabs/puppetlabs-docker/runs/6862356563?check_suite_focus=true

chelnak:
Hey @maltewhiite - I'm not sure on the state of this PR.

The associated issue states that `create` doesn't fix the problem. I'm also not sure that suggested implementation is the correct route for a fix.

Thanks!

chelnak:
Hey @maltewhiite, I'm going to close this PR for now. As I mentioned in the previous comment, I'm not sure if the suggested implementation is the right way to attack this issue.

If you have any other ideas, please do open another PR.

In the meantime, I'd like thank you for your contribution!

maltewhiite:
@chelnak all good. My team made a local fix to the problem.

chelnak:
@maltewhiite You too! Thanks again for the contribution.

PR #846, opened 2022-07-04
adding optional variable for package_key_check_source to RedHat

STaegtmeier:
Change the hard coded value for `$package_key_check_source` for RedHat to the optional variable `$docker_package_key_check_source`.
The variable is already used further down in the following else-block for RedHat.
The optional var is defined as `true` within the `params.pp` so there should be no following issues.

The change is needed for one of my customers which used an own yum-repo for the docker repository, were the values must be set to `false`.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -588,7 +588,7 @@ class docker (
         'RedHat' : {
           $package_location         = $docker_ce_source_location
           $package_key_source       = $docker_ce_key_source
-          $package_key_check_source = true
+          $package_key_check_source = $docker_package_key_check_source
         }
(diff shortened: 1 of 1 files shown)


chelnak:
@STaegtmeier Thanks for the contribution. 

Should the following line also consume the `docker_package_key_check_source` variable?

https://github.com/puppetlabs/puppetlabs-docker/blob/30d37d6cde66a5a66626543592f9cdde41e505ea/manifests/init.pp#L573

chelnak (approved):
Thanks for making the change. This looks fine now.

PR #852, opened 2022-08-12
Make management of systemd override directories optional

antaflos:
When using puppet-systemd and the `systemd::dropin_file` type to manage
custom Docker-related overrides the puppet-systemd module takes care of
creating and managing the required override directories like
/etc/systemd/system/docker.service.d and
/etc/systemd/system/docker.socket.d. This leads to duplicate resource
declaration errors because the puppetlabs-docker module also wants to
manage these systemd override directories.

This change adds a boolean parameter `manage_systemd_override_dirs` to
control whether this module should manage the directories
/etc/systemd/system/docker.service.d and
/etc/systemd/system/docker.socket.d. If set to 'false' then these
directories will not be touched by this module and the user must take
care of managing them, e.g. by means of puppet-systemd and
`systemd::dropin_file`. The parameter `manage_systemd_override_dirs` is
set to 'true' by default, thus backwards-compatibility is preserved.

Contains updated spec tests and parameter documentation.

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -33,6 +33,18 @@
 #   Valid values are 'true', 'false'.
 #   Defaults to 'true'.
 #
+# @param manage_systemd_override_dirs
+#   Specify whether the systemd override directories such as
+#   /etc/systemd/systemd/docker.service.d and
+#   /etc/systemd/system/docker.socket.d should be managed by this module or
+#   not. Set this to 'false' if you manage custom Docker-related systemd
+#   overrides, for example with `systemd::dropin_file` from puppet-systemd.
+#   Otherwise this will lead to duplicate resource declaration errors because
+#   both this module and puppet-systemd want to manage the override
+#   directories.
+#   Valid values are 'true', 'false'.
+#   Defaults to 'true'.
+#
 # @param docker_command
 #
 # @param docker_start_command
@@ -204,6 +216,7 @@ class docker::service (
   $service_state                     = $docker::service_state,
   $service_enable                    = $docker::service_enable,
   $manage_service                    = $docker::manage_service,
+  $manage_systemd_override_dirs      = $docker::manage_systemd_override_dirs,
   $root_dir                          = $docker::root_dir,
   $extra_parameters                  = $docker::extra_parameters,
   $shell_values                      = $docker::shell_values,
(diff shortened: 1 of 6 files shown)


manifests/service.pp
```diff
@@ -312,8 +325,10 @@
 
   case $service_provider {
     'systemd': {
-      file { '/etc/systemd/system/docker.service.d':
-        ensure => 'directory',
+      if $manage_systemd_override_dirs {
```
bastelfreak, line 328: I think those directories should not be created here at all  instead the puppet/systemd module should be used

antaflos: I agree, puppetlabs-docker should make use of the facilities provided by puppet-systemd in such cases. I was hesitant to propose or implement that because it is probably a larger change, and adds a new dependency.

antaflos:
The `Windows 2022, puppet6-nightly` check fails because of network timeout issues, certainly unrelated to the changes proposed in this PR. See https://github.com/puppetlabs/puppetlabs-docker/runs/8186444366?check_suite_focus=true#step:13:73 for example. Not sure how or if I can trigger running the tests again.

manifests/service.pp
bastelfreak: personally I would highly prefer that. maybe wait for feedback from @modules. I think the current implementation is really against best practices and adding the manage_systemd_override_dirs just makes it more challenging in the future to switch to puppet/systemd module without breaking changes

antaflos:
Bumping for visibility, possibly.

pmcmaw:
Hey @antaflos thanks for that ping! 
Your codebase is out of date, would you mind rebasing?
I will get this on the list to be reviewed at our next community day (runs most Mondays)

antaflos:
Thanks @pmcmaw, I just rebased.

antaflos:
Just rebased again to keep up to date. Any movement regarding this PR?

manifests/service.pp
GSPatton: Hi @antaflos and @bastelfreak. The Docker module leveraging puppet-systemd sounds like a sensible approach and we would happy to see a PR implementing such a change.

LukasAud:
Hi @antaflos, we have noticed this PR has been inactive for a while. An old review above by @bastelfreak suggested an alternative approach to this issue and we have agreed that it was a better solution to this problem. As such, this PR should be closed and an issue requesting that feature opened in our Issue tab here at GitHub. Thanks for your understanding.

jordanbreen28:
Closing inline with comments above.

PR #856, opened 2022-09-07
Update supported Ubuntu versions in README

nikzeyn:
Update the Ubuntu versions that this module supports. Currently, it mentions that it supports Ubuntu 14.04 and Ubuntu 16.04. It seems like they are no longer supported.
According to puppetlabs-docker/metadata.json file, the supported versions are: 18.04, 20.04 and 22.04.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1016,9 +1016,9 @@ This module supports:
 * Debian 9.0
 * Debian 10
 * RedHat 7.0 - limited support available
-* Ubuntu 14.04
-* Ubuntu 16.04
 * Ubuntu 18.04
+* Ubuntu 20.04
+* Ubuntu 22.04
 * Windows Server 2016 (Docker Enterprise Edition only)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1016,9 +1016,9 @@ This module supports:
 * Debian 9.0
 * Debian 10
```
smortex, line 1017:
```suggestion
* Debian 10
* Debian 11
```

README.md
```diff
@@ -1016,9 +1016,9 @@ This module supports:
...
 * RedHat 7.0 - limited support available
-* Ubuntu 14.04
-* Ubuntu 16.04
 * Ubuntu 18.04
+* Ubuntu 20.04
+* Ubuntu 22.04
 * Windows Server 2016 (Docker Enterprise Edition only)
```
smortex, line 1022:
```suggestion
* Windows Server 2016 (Docker Enterprise Edition only)
* Windows Server 2019 (Docker Enterprise Edition only)
* Windows Server 2022 (Docker Enterprise Edition only)
```

smortex (commented):
Not sure if we should fix this list or just remove it so that it does not become deprecated over time. If we choose to keep it, please update the other supported OS too: I added in-line comments).

nikzeyn:
> Not sure if we should fix this list or just remove it so that it does not become deprecated over time. If we choose to keep it, please update the other supported OS too: I added in-line comments).

I think it's good to keep the list of supported systems.

smortex (approved):
LGTM

GSPatton (approved):
This looks good. We are updating all of the supported OSes on our modules soon so thanks for getting us started.

Debian 8 isn't supported, but I'll merge anyway and if you feel like raising another PR for it, go for it. Otherwise we'll sort that soon.

PR #858, opened 2022-09-16
Change `stop_wait_time` value to match Docker default

sebcdri:
According to Docker documentation, [the `--time` option default value is  10 seconds](https://docs.docker.com/engine/reference/commandline/stop/).

The module set this value to 0, causing immediate destruction of the container instead of leaving it a 10s grace time to shutdown.

Use the default value of 10 to honor the default behavior.


neomilium (approved):

LukasAud (approved):

LukasAud:
LGTM. Failures are unrelated. Thanks for your contribution!

PR #863, opened 2022-10-12
(maint) Hardening manifests and tasks

LukasAud:
This PR aims to implement some changes that ensure no malformed commands are passed through to the system. Certain commands were left undivided as the commands did not get correctly interpreted.

Primarily, the commands targeted were the ones related to Open3 and exec.

### DIFF
--- a/manifests/compose.pp
+++ b/manifests/compose.pp
@@ -71,7 +71,7 @@ class docker::compose (
     }
 
     if $facts['os']['family'] == 'windows' {
-      $docker_download_command = "if (Invoke-WebRequest ${docker_compose_url} ${proxy_opt} -UseBasicParsing -OutFile \"${docker_compose_location_versioned}\") { exit 0 } else { exit 1}" # lint:ignore:140chars
+      $docker_download_command = ['if', '(Invoke-WebRequest', $docker_compose_url, $proxy_opt, '-UseBasicParsing', '-OutFile', "\"${docker_compose_location_versioned}\") { exit 0 } else { exit 1}"] # lint:ignore:140chars
 
       exec { "Install Docker Compose ${version}":
         command  => template('docker/windows/download_docker_compose.ps1.erb'),
@@ -92,7 +92,7 @@ class docker::compose (
       exec { "Install Docker Compose ${version}":
         path    => '/usr/bin/',
         cwd     => '/tmp',
-        command => "curl -s -S -L ${proxy_opt} ${docker_compose_url} -o ${docker_compose_location_versioned}",
+        command => ['curl', '-s', '-S', '-L', $proxy_opt, $docker_compose_url, '-o', $docker_compose_location_versioned],
         creates => $docker_compose_location_versioned,
         require => Package['curl'],
       }
--- a/manifests/compose.pp
+++ b/manifests/compose.pp
@@ -89,10 +89,11 @@ class docker::compose (
         ensure_packages(['curl'])
       }
 
+      $compose_install = "curl -s -S -L ${proxy_opt} ${docker_compose_url} -o ${docker_compose_location_versioned}"
       exec { "Install Docker Compose ${version}":
         path    => '/usr/bin/',
         cwd     => '/tmp',
-        command => ['curl', '-s', '-S', '-L', $proxy_opt, $docker_compose_url, '-o', $docker_compose_location_versioned],
+        command => $compose_install,
         creates => $docker_compose_location_versioned,
         require => Package['curl'],
       }
--- a/manifests/compose.pp
+++ b/manifests/compose.pp
@@ -71,8 +71,6 @@ class docker::compose (
     }
 
     if $facts['os']['family'] == 'windows' {
-      $docker_download_command = ['if', '(Invoke-WebRequest', $docker_compose_url, $proxy_opt, '-UseBasicParsing', '-OutFile', "\"${docker_compose_location_versioned}\") { exit 0 } else { exit 1}"] # lint:ignore:140chars
-
       exec { "Install Docker Compose ${version}":
         command  => template('docker/windows/download_docker_compose.ps1.erb'),
         provider => powershell,
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -113,8 +113,9 @@ class docker::install (
             }
           }
 
+          $fail_restart_command = ['SC.exe', 'failure', 'Docker', 'reset= 432000', 'actions= restart/30000/restart/60000/restart/60000']
           exec { 'service-restart-on-failure':
-            command     => 'SC.exe failure Docker reset= 432000 actions= restart/30000/restart/60000/restart/60000',
+            command     => $fail_restart_command,
             refreshonly => true,
             logoutput   => true,
             provider    => powershell,
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -113,7 +113,7 @@ class docker::install (
             }
           }
 
-          $fail_restart_command = ['SC.exe', 'failure', 'Docker', 'reset= 432000', 'actions= restart/30000/restart/60000/restart/60000']
+          $fail_restart_command = 'SC.exe failure Docker reset= 432000 actions= restart/30000/restart/60000/restart/60000'
           exec { 'service-restart-on-failure':
             command     => $fail_restart_command,
             refreshonly => true,
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -113,9 +113,8 @@ class docker::install (
             }
           }
 
-          $fail_restart_command = 'SC.exe failure Docker reset= 432000 actions= restart/30000/restart/60000/restart/60000'
           exec { 'service-restart-on-failure':
-            command     => $fail_restart_command,
+            command     => 'SC.exe failure Docker reset= 432000 actions= restart/30000/restart/60000/restart/60000',
             refreshonly => true,
             logoutput   => true,
             provider    => powershell,
--- a/manifests/machine.pp
+++ b/manifests/machine.pp
@@ -56,8 +56,6 @@ class docker::machine (
     }
 
     if $facts['os']['family'] == 'windows' {
-      $docker_download_command = ['if', '(Invoke-WebRequest', $docker_machine_url, $proxy_opt, '-UseBasicParsing', '-OutFile', "\"${docker_machine_location_versioned}\") { exit 0 } else { exit 1}"] # lint:ignore:140chars
-
       exec { "Install Docker Machine ${version}":
         command  => template('docker/windows/download_docker_machine.ps1.erb'),
         provider => powershell,
--- a/manifests/machine.pp
+++ b/manifests/machine.pp
@@ -56,7 +56,7 @@ class docker::machine (
     }
 
     if $facts['os']['family'] == 'windows' {
-      $docker_download_command = "if (Invoke-WebRequest ${docker_machine_url} ${proxy_opt} -UseBasicParsing -OutFile \"${docker_machine_location_versioned}\") { exit 0 } else { exit 1}" # lint:ignore:140chars
+      $docker_download_command = ['if', '(Invoke-WebRequest', $docker_machine_url, $proxy_opt, '-UseBasicParsing', '-OutFile', "\"${docker_machine_location_versioned}\") { exit 0 } else { exit 1}"] # lint:ignore:140chars
 
       exec { "Install Docker Machine ${version}":
         command  => template('docker/windows/download_docker_machine.ps1.erb'),
@@ -74,10 +74,11 @@ class docker::machine (
         ensure_packages(['curl'])
       }
 
+      $install_command = ['curl', '-s', '-S', '-L', $proxy_opt, $docker_machine_url, '-o', $docker_machine_location_versioned] # lint:ignore:140chars
       exec { "Install Docker Machine ${version}":
         path    => '/usr/bin/',
         cwd     => '/tmp',
-        command => "curl -s -S -L ${proxy_opt} ${docker_machine_url} -o ${docker_machine_location_versioned}",
+        command => $install_command,
         creates => $docker_machine_location_versioned,
         require => Package['curl'],
       }
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -68,8 +68,8 @@ define docker::plugin (
       }
     )
 
-    $exec_install   = [$docker_command, 'install', $docker_plugin_install_flags]
-    $unless_install = [$docker_command, 'ls', "--format='{{.PluginReference}}' | grep -w ${plugin_name}"]
+    $exec_install   = "${docker_command} install ${docker_plugin_install_flags}"
+    $unless_install = "${docker_command} ls --format='{{.PluginReference}}' | grep -w ${plugin_name}"
 
     exec { "plugin install ${plugin_name}":
       command     => $exec_install,
@@ -86,7 +86,7 @@ define docker::plugin (
     )
 
     $exec_rm   = [$docker_command, 'rm', $docker_plugin_remove_flags]
-    $onlyif_rm = [$docker_command, 'ls', "--format='{{.PluginReference}}' | grep -w ${plugin_name}"]
+    $onlyif_rm = "${docker_command} ls --format='{{.PluginReference}}' | grep -w ${plugin_name}"
 
     exec { "plugin remove ${plugin_name}":
       command     => $exec_rm,
@@ -106,7 +106,7 @@ define docker::plugin (
     )
 
     $exec_enable   = [$docker_command, 'enable', $docker_plugin_enable_flags]
-    $onlyif_enable = [$docker_command, 'ls', '-f', "enabled=false --format='{{.PluginReference}}' | grep -w ${plugin_name}"]
+    $onlyif_enable = "${docker_command} ls -f enabled=false --format='{{.PluginReference}}' | grep -w ${plugin_name}"
 
     exec { "plugin enable ${plugin_name}":
       command     => $exec_enable,
@@ -117,7 +117,7 @@ define docker::plugin (
     }
   } elsif $enabled == false {
     $else_command = [$docker_command, 'disable', $plugin_name]
-    $else_unless = [$docker_command, 'ls', '-f', "enabled=false --format='{{.PluginReference}}' | grep -w ${plugin_name}"]
+    $else_unless = "${docker_command} ls -f enabled=false --format='{{.PluginReference}}' | grep -w ${plugin_name}"
 
     exec { "disable ${plugin_name}":
       command     => $else_command,
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -68,8 +68,8 @@ define docker::plugin (
       }
     )
 
-    $exec_install   = "${docker_command} install ${docker_plugin_install_flags}"
-    $unless_install = "${docker_command} ls --format='{{.PluginReference}}' | grep -w ${plugin_name}"
+    $exec_install   = [$docker_command, 'install', $docker_plugin_install_flags]
+    $unless_install = [$docker_command, 'ls', "--format='{{.PluginReference}}' | grep -w ${plugin_name}"]
 
     exec { "plugin install ${plugin_name}":
       command     => $exec_install,
@@ -85,8 +85,8 @@ define docker::plugin (
       }
     )
 
-    $exec_rm   = "${docker_command} rm ${docker_plugin_remove_flags}"
-    $onlyif_rm = "${docker_command} ls --format='{{.PluginReference}}' | grep -w ${plugin_name}"
+    $exec_rm   = [$docker_command, 'rm', $docker_plugin_remove_flags]
+    $onlyif_rm = [$docker_command, 'ls', "--format='{{.PluginReference}}' | grep -w ${plugin_name}"]
 
     exec { "plugin remove ${plugin_name}":
       command     => $exec_rm,
@@ -105,8 +105,8 @@ define docker::plugin (
       }
     )
 
-    $exec_enable   = "${docker_command} enable ${docker_plugin_enable_flags}"
-    $onlyif_enable = "${docker_command} ls -f enabled=false --format='{{.PluginReference}}' | grep -w ${plugin_name}"
+    $exec_enable   = [$docker_command, 'enable', $docker_plugin_enable_flags]
+    $onlyif_enable = [$docker_command, 'ls', '-f', "enabled=false --format='{{.PluginReference}}' | grep -w ${plugin_name}"]
 
     exec { "plugin enable ${plugin_name}":
       command     => $exec_enable,
@@ -116,12 +116,15 @@ define docker::plugin (
       onlyif      => $onlyif_enable,
     }
   } elsif $enabled == false {
+    $else_command = [$docker_command, 'disable', $plugin_name]
+    $else_unless = [$docker_command, 'ls', '-f', "enabled=false --format='{{.PluginReference}}' | grep -w ${plugin_name}"]
+
     exec { "disable ${plugin_name}":
-      command     => "${docker_command} disable ${plugin_name}",
+      command     => $else_command,
       environment => 'HOME=/root',
       path        => ['/bin', '/usr/bin',],
       timeout     => 0,
-      unless      => "${docker_command} ls -f enabled=false --format='{{.PluginReference}}' | grep -w ${plugin_name}",
+      unless      => $else_unless,
     }
   }
 }
--- a/manifests/registry.pp
+++ b/manifests/registry.pp
@@ -125,7 +125,7 @@ define docker::registry (
       # server may be an URI, which can contain /
       $server_strip  = regsubst($server, '[/:]', '_', 'G')
       $passfile      = "${::docker_user_temp_path}/registry-auth-puppet_receipt_${server_strip}_${local_user}"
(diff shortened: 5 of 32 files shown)


manifests/compose.pp
```diff
@@ -71,7 +71,7 @@
     }
 
     if $facts['os']['family'] == 'windows' {
-      $docker_download_command = "if (Invoke-WebRequest ${docker_compose_url} ${proxy_opt} -UseBasicParsing -OutFile \"${docker_compose_location_versioned}\") { exit 0 } else { exit 1}" # lint:ignore:140chars
+      $docker_download_command = ['if', '(Invoke-WebRequest', $docker_compose_url, $proxy_opt, '-UseBasicParsing', '-OutFile', "\"${docker_compose_location_versioned}\") { exit 0 } else { exit 1}"] # lint:ignore:140chars
```
chelnak, line 74: I believe that this variable is unused so we could potentially remove it?

manifests/compose.pp
```diff
@@ -89,10 +89,11 @@
         ensure_packages(['curl'])
       }
 
+      $compose_install = "curl -s -S -L ${proxy_opt} ${docker_compose_url} -o ${docker_compose_location_versioned}"
```
chelnak, line 92: Should this not be broken in to args?

```suggestion
      $compose_install = ['curl', '-s', '-S', '-L', $proxy_opt, docker_compose_url, '-o', $docker_compose_location_versioned]"
```

manifests/install.pp
```diff
@@ -113,8 +113,9 @@
             }
           }
 
+          $fail_restart_command = 'SC.exe failure Docker reset= 432000 actions= restart/30000/restart/60000/restart/60000'
```
chelnak, line 116: This is potentially an unneeded addition given that there is no interpolation.

manifests/machine.pp
```diff
@@ -56,7 +56,7 @@
     }
 
     if $facts['os']['family'] == 'windows' {
-      $docker_download_command = "if (Invoke-WebRequest ${docker_machine_url} ${proxy_opt} -UseBasicParsing -OutFile \"${docker_machine_location_versioned}\") { exit 0 } else { exit 1}" # lint:ignore:140chars
+      $docker_download_command = ['if', '(Invoke-WebRequest', $docker_machine_url, $proxy_opt, '-UseBasicParsing', '-OutFile', "\"${docker_machine_location_versioned}\") { exit 0 } else { exit 1}"] # lint:ignore:140chars
```
chelnak, line 59: I believe that this variable is unused so we could potentially remove it?

manifests/plugin.pp
```diff
@@ -79,13 +79,13 @@
...
-    $docker_plugin_remove_flags = docker_plugin_remove_flags( {
+    $docker_plugin_remove_flags = docker_plugin_remove_flags({
         plugin_name => $plugin_name,
         force_remove => $force_remove,
       }
     )
 
-    $exec_rm   = "${docker_command} rm ${docker_plugin_remove_flags}"
+    $exec_rm   = [$docker_command, 'rm', $docker_plugin_remove_flags]
     $onlyif_rm = "${docker_command} ls --format='{{.PluginReference}}' | grep -w ${plugin_name}"
```
chelnak, line 89: This is an interesting one. `$onlyif_rm` should be broken out because both `$docker_command` and `$plugin_name` come from parameters.

With exec, both `onlyif` and `unless` accept a `String` or an `Array[Array[String]]`.

For the latter it would look like this

```
onlyif => [ ['my', 'command', 'and', 'args'] ]

```

When it is evaluated exec looks to see if all commands are truthy (e.g you may have multiple inner arrays for different commands).

The tricky part here is that when using an array of args, exec will disregard any metacharacters (right word??). In this case it would ignore the pipe and everything to the right of it.

There may be an example of how to handle this in apt or mysql. Alternatively we could just clean the strings with shell_escape or something similar.

manifests/registry.pp
```diff
@@ -125,7 +125,7 @@
       # server may be an URI, which can contain /
       $server_strip  = regsubst($server, '[/:]', '_', 'G')
       $passfile      = "${::docker_user_temp_path}/registry-auth-puppet_receipt_${server_strip}_${local_user}"
-      $_auth_command = "if (-not (${auth_cmd})) { Remove-Item -Path ${passfile} -Force -Recurse -EA SilentlyContinue; exit 1 } else { exit 0 }" # lint:ignore:140chars
+      $_auth_command = ["if (-not (${auth_cmd}))", "{ Remove-Item -Path ${passfile}", '-Force', '-Recurse', '-EA', 'SilentlyContinue; exit 1 } else { exit 0 }'] # lint:ignore:140chars
```
chelnak, line 128: There is another variant of this on line 115 that needs addressing.

Additionally I'm not 100% about having logic like this in an exec command.

Can it be refactored to use onlyif/unless or even broken out in to a script template?

manifests/run.pp
```diff
@@ -385,18 +387,24 @@
 
   if $restart {
     if $ensure == 'absent' {
+      $restart_stop_command = [$docker_command, 'stop', "--time=${stop_wait_time}", $sanitised_title]
```
chelnak, line 390: `"--time=${stop_wait_time}"`

This will need addressing.

Does the flag actually need the `=`? if not you could just have a clean break between `--time` and `$stop_wait_time`.

manifests/run.pp
```diff
@@ -385,18 +387,24 @@
...
-        onlyif      => "${docker_command} inspect ${sanitised_title}",
+        command     => $restart_stop_command,
+        onlyif      => $restart_stop_onlyif,
         environment => $exec_environment,
         path        => $exec_path,
         provider    => $exec_provider,
         timeout     => $exec_timeout,
       }
 
+      $restart_remove_command = "${docker_command} rm -v ${sanitised_title}"
```
chelnak, line 402: This command should be broken down in to an array of args.

manifests/run.pp
```diff
@@ -385,18 +387,24 @@
 
   if $restart {
     if $ensure == 'absent' {
+      $restart_stop_command = [$docker_command, 'stop', "--time=${stop_wait_time}", $sanitised_title]
+      $restart_stop_onlyif = [$docker_command, 'inspect', $sanitised_title]
```
chelnak, line 391: I don't think this will work as expected. `onlyif` will be expecting an array of arrays.

```suggestion
      $restart_stop_onlyif = [[$docker_command, 'inspect', $sanitised_title][
```

manifests/run.pp
```diff
@@ -385,18 +387,24 @@
...
+        command     => $restart_stop_command,
+        onlyif      => $restart_stop_onlyif,
         environment => $exec_environment,
         path        => $exec_path,
         provider    => $exec_provider,
         timeout     => $exec_timeout,
       }
 
+      $restart_remove_command = "${docker_command} rm -v ${sanitised_title}"
+      $restart_remove_onlyif = [$docker_command, 'inspect', $sanitised_title]
```
chelnak, line 403: I don't think this will work as expected. onlyif will be expecting an array of arrays.

```suggestion
      $restart_remove_onlyif = [[$docker_command, 'inspect', $sanitised_title]]
```

manifests/run.pp
```diff
@@ -432,17 +440,19 @@
         }
 
         if $running == false {
+          $running_stop_command = [$docker_command, 'stop', "--time=${stop_wait_time}", $sanitised_title]
```
chelnak, line 443: "--time=${stop_wait_time}"

This will need addressing.

Does the flag actually need the =? if not you could just have a clean break between --time and $stop_wait_time.

manifests/run.pp
```diff
@@ -517,9 +527,12 @@
 
     if $ensure == 'absent' {
       if $facts['os']['family'] == 'windows' {
+        $absent_stop_command = "${docker_command} stop --time=${stop_wait_time} ${sanitised_title}"
+        $absent_stop_onlyif = "${docker_command} inspect ${sanitised_title}"
```
chelnak, line 531: Both of these need breaking up.

Also refer to the above comment for the `--time` arg.

tasks/node_rm.rb
```diff
@@ -6,9 +6,9 @@
 require 'puppet'
 
 def node_rm(force, node)
-  cmd_string = 'docker node rm'
-  cmd_string += ' --force' unless force.nil?
-  cmd_string += " #{node}" unless node.nil?
+  cmd_string = ['docker', 'node', 'rm']
+  cmd_string += [' --force'] unless force.nil?
+  cmd_string += [" #{node}"] unless node.nil?
```
chelnak, line 11: We should try to avoid using += here.

I believe [Array#concat](https://ruby-doc.org/core-2.1.1/Array.html#method-i-concat) would be preferable in this instance.

Also lets rename the variable to cmd rather than cmd_string.

tasks/node_update.rb
```diff
@@ -6,19 +6,19 @@
...
-  cmd_string += " --availability #{availability}" unless availability.nil?
-  cmd_string += " --role #{role}" unless role.nil?
+  cmd_string = ['docker', 'node', 'update']
+  cmd_string += [" --availability #{availability}"] unless availability.nil?
+  cmd_string += [" --role #{role}"] unless role.nil?
 
   if label_add.is_a? Array
     label_add.each do |param|
-      cmd_string += " --label-add #{param}"
+      cmd_string += [" --label-add #{param}"]
```
chelnak, line 15: See above comment

tasks/service_create.rb
```diff
@@ -6,18 +6,18 @@
 require 'puppet'
 
 def service_create(image, replicas, expose, env, command, extra_params, service, detach)
-  cmd_string = 'docker service create'
+  cmd_string = ['docker', 'service', 'create']
```
chelnak, line 9: We should try to avoid using += here.

I believe [Array#concat](https://ruby-doc.org/core-2.1.1/Array.html#method-i-concat) would be preferable in this instance.

This applies to all instances in this file.

Also lets rename the variable to cmd rather than cmd_string.

tasks/service_rm.rb
```diff
@@ -6,8 +6,8 @@
 require 'puppet'
 
 def service_rm(service)
-  cmd_string = 'docker service rm'
-  cmd_string += " #{service}" unless service.nil?
+  cmd_string = ['docker', 'service', 'rm']
+  cmd_string += [" #{service}"] unless service.nil?
```
chelnak, line 10: We should try to avoid using += here.

I believe [Array#concat](https://ruby-doc.org/core-2.1.1/Array.html#method-i-concat) would be preferable in this instance.

Also lets rename the variable to cmd rather than cmd_string.

tasks/service_scale.rb
```diff
@@ -6,10 +6,10 @@
...
 
 def service_scale(service, scale, detach)
-  cmd_string = 'docker service scale'
-  cmd_string += " #{service}" unless service.nil?
-  cmd_string += "=#{scale}" unless scale.nil?
-  cmd_string += ' -d' unless detach.nil?
+  cmd_string = ['docker', 'service', 'scale']
+  cmd_string += [" #{service}"] unless service.nil?
+  cmd_string += ["=#{scale}"] unless scale.nil?
+  cmd_string += [' -d'] unless detach.nil?
```
chelnak, line 12: We should try to avoid using += here.

I believe [Array#concat](https://ruby-doc.org/core-2.1.1/Array.html#method-i-concat) would be preferable in this instance.

Also lets rename the variable to cmd rather than cmd_string.

tasks/swarm_init.rb
```diff
@@ -6,16 +6,16 @@
...
+  cmd_string = ['docker', 'swarm', 'init']
+  cmd_string += [" --advertise-addr=#{advertise_addr}"] unless advertise_addr.nil?
+  cmd_string += [' --autolock'] unless autolock.nil?
+  cmd_string += [' --cert-expiry'] unless cert_expiry.nil?
+  cmd_string += [" --dispatcher-heartbeat=#{dispatcher_heartbeat}"] unless dispatcher_heartbeat.nil?
+  cmd_string += [" --external-ca=#{external_ca}"] unless external_ca.nil?
+  cmd_string += [' --force-new-cluster'] unless force_new_cluster.nil?
+  cmd_string += [" --listen-addr=#{listen_addr}"] unless listen_addr.nil?
+  cmd_string += [" --max-snapshots=#{max_snapshots}"] unless max_snapshots.nil?
+  cmd_string += [" --snapshot-interval=#{snapshot_interval}"] unless snapshot_interval.nil?
```
chelnak, line 18: We should try to avoid using += here.

I believe [Array#concat](https://ruby-doc.org/core-2.1.1/Array.html#method-i-concat) would be preferable in this instance.

Also lets rename the variable to cmd rather than cmd_string.

tasks/swarm_join.rb
```diff
@@ -6,11 +6,11 @@
...
-  cmd_string = 'docker swarm join'
-  cmd_string += " --advertise-addr=#{advertise_addr}" unless advertise_addr.nil?
-  cmd_string += " --listen-addr=#{listen_addr}" unless listen_addr.nil?
-  cmd_string += " --token=#{token}" unless token.nil?
-  cmd_string += " #{manager_ip}" unless manager_ip.nil?
+  cmd_string = ['docker', 'swarm', 'join']
+  cmd_string += [" --advertise-addr=#{advertise_addr}"] unless advertise_addr.nil?
+  cmd_string += [" --listen-addr=#{listen_addr}"] unless listen_addr.nil?
+  cmd_string += [" --token=#{token}"] unless token.nil?
+  cmd_string += [" #{manager_ip}"] unless manager_ip.nil?
```
chelnak, line 13: We should try to avoid using += here.

I believe [Array#concat](https://ruby-doc.org/core-2.1.1/Array.html#method-i-concat) would be preferable in this instance.

Also lets rename the variable to cmd rather than cmd_string.

tasks/swarm_leave.rb
```diff
@@ -6,8 +6,8 @@
 require 'puppet'
 
 def swarm_leave(force)
-  cmd_string = 'docker swarm leave '
-  cmd_string += ' -f' if force == 'true'
+  cmd_string = ['docker', 'swarm', 'leave']
+  cmd_string += [' -f'] if force == 'true'
```
chelnak, line 10: We should try to avoid using += here.

I believe [Array#concat](https://ruby-doc.org/core-2.1.1/Array.html#method-i-concat) would be preferable in this instance.

Also lets rename the variable to cmd rather than cmd_string.

tasks/swarm_token.rb
```diff
@@ -6,8 +6,8 @@
 require 'puppet'
 
 def swarm_token(node_role)
-  cmd_string = 'docker swarm join-token -q'
-  cmd_string += " #{node_role}" unless node_role.nil?
+  cmd_string = ['docker', 'swarm', 'join-token', '-q']
+  cmd_string += [" #{node_role}"] unless node_role.nil?
```
chelnak, line 10: We should try to avoid using += here.

I believe [Array#concat](https://ruby-doc.org/core-2.1.1/Array.html#method-i-concat) would be preferable in this instance.

Also lets rename the variable to cmd rather than cmd_string.

chelnak (changes requested):
Great work @LukasAud !

I've left some comments for you to address.

manifests/compose.pp
LukasAud: Anything that is not broken into args will cry, scream and kick at me when I run test cases. Tried to figure out the reason behind those failures but eventually had to give up and just let some commands stay the same.

manifests/install.pp
LukasAud: Whoops, must have slipped while I was in zombie mode working on those commands. Will undo.

manifests/plugin.pp
LukasAud: Talked with @david22swan about this cases. I believe we just agreed to not touch anything with pipes because of the reason mentioned. I did try some variations arraying and segmenting like:
`$onlyif_rm = ["$docker_command, 'ls', "--format='{{.PluginReference}}' | grep -w ${plugin_name}"]`

Just for the sake of testing whether the problem was separating the pipe from the rest of the arguments but it did not work. Currently any command with a grep is untouched in this and every other PR afaik.

manifests/run.pp
LukasAud: Tried to keep the commands as untouched as I could. After all, this is not a performance update or code cleanup PR. I can still try to modify it if you think its worth the effort.

manifests/run.pp
LukasAud: Same as before. If its not broken already, then it throws some serious issues when running tests.

manifests/run.pp
LukasAud: Did not have any problems defining it as a simple array tbh. And that is pretty weird for a module that throws me 100s of failures for a single misplaced comma. Will try to implement tho.

manifests/run.pp
LukasAud: Same as above. Will try to implement.

manifests/run.pp
LukasAud: Same as above.

manifests/run.pp
LukasAud: Same as above. If not broken up, it presents serious issues. To be more precise, I think most of the unbroken arguments in run.pp were breaking windows acceptance tests when tried to modify them. I cannot really tell you why. I dont feel like there is any reason windows should not be able to handle a 
`[$docker_command, 'inspect', $sanitised_title]`

manifests/run.pp
```diff
@@ -536,9 +549,12 @@
           notify    => Exec["remove container ${service_prefix}${sanitised_title}"],
         }
       }
+      $absent_remove_command = "${docker_command} rm -v ${sanitised_title}"
+      $absent_remove_onlyif = "${docker_command} inspect ${sanitised_title}"
```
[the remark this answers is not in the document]
LukasAud: same as above

spec/shared_examples/services.rb
```diff
@@ -108,7 +108,7 @@
     #   'extra_params' => Array(extra_params),
     # )
 
-    exec_scale = "#{docker_command} scale #{service_name}=#{replicas}"
+    exec_scale = [docker_command, 'scale', "#{service_name}=#{replicas}"]
```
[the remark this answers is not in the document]
LukasAud: This is a spec test case, not sure how would that be done. Nor why.

tasks/node_rm.rb
LukasAud: See above.

tasks/node_update.rb
LukasAud: See above.

tasks/node_update.rb
```diff
@@ -6,19 +6,19 @@
...
     end
   end
 
   if label_rm.is_a? Array
     label_rm.each do |param|
-      cmd_string += " --label-rm #{param}"
+      cmd_string += [" --label-rm #{param}"]
```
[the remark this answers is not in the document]
LukasAud: See above.

tasks/service_create.rb
LukasAud: See above.

tasks/service_rm.rb
LukasAud: See above.

tasks/service_update.rb
```diff
@@ -6,18 +6,18 @@
 require 'puppet'
 
 def service_update(image, service, constraint_add, constraint_rm)
-  cmd_string = 'docker service update'
-  cmd_string += " --image #{image}" unless image.nil?
+  cmd_string = ['docker', 'service', 'update']
```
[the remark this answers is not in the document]
LukasAud: See above.

tasks/swarm_init.rb
LukasAud: See above.

tasks/swarm_join.rb
LukasAud: See above.

tasks/swarm_leave.rb
LukasAud: See above.

tasks/swarm_token.rb
LukasAud: See above.

manifests/compose.pp
LukasAud: Will try to implement but wont know if it works until it hits github.

manifests/run.pp
chelnak: I think it's worth the effort yes.

The suggestion there is simply suggestion a possible way to make it work without interpolation.

manifests/run.pp
chelnak: This should be broken out and tests updated accordingly.

manifests/run.pp
chelnak: It would be good to spend a little bit of time understanding why the failures are occurring. 

The commands wont reach the os as an array. They are stringified by the provider.

spec/shared_examples/services.rb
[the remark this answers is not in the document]
chelnak: It should probably reflect the implementation to be honest.

tasks/node_rm.rb
chelnak: In this case you have changed from concatenating a string to an array so we should generally lean towards the more idiomatic approach.

tasks/node_update.rb
```diff
@@ -6,19 +6,19 @@
 require 'puppet'
 
 def node_update(availability, role, label_add, label_rm, node)
-  cmd_string = 'docker node update'
-  cmd_string += " --availability #{availability}" unless availability.nil?
-  cmd_string += " --role #{role}" unless role.nil?
+  cmd_string = ['docker', 'node', 'update']
+  cmd_string += [" --availability #{availability}"] unless availability.nil?
+  cmd_string += [" --role #{role}"] unless role.nil?
```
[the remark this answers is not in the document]
chelnak: In this case you have changed from concatenating a string to an array so we should generally lean towards the more idiomatic approach.

tasks/node_update.rb
chelnak: In this case you have changed from concatenating a string to an array so we should generally lean towards the more idiomatic approach.

tasks/service_create.rb
chelnak: In this case you have changed from concatenating a string to an array so we should generally lean towards the more idiomatic approach.

tasks/service_rm.rb
chelnak: In this case you have changed from concatenating a string to an array so we should generally lean towards the more idiomatic approach.

tasks/service_scale.rb
chelnak: In this case you have changed from concatenating a string to an array so we should generally lean towards the more idiomatic approach.

tasks/service_update.rb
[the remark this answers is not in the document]
chelnak: In this case you have changed from concatenating a string to an array so we should generally lean towards the more idiomatic approach.

tasks/swarm_init.rb
chelnak: In this case you have changed from concatenating a string to an array so we should generally lean towards the more idiomatic approach.

tasks/swarm_join.rb
chelnak: In this case you have changed from concatenating a string to an array so we should generally lean towards the more idiomatic approach.

tasks/swarm_leave.rb
chelnak: In this case you have changed from concatenating a string to an array so we should generally lean towards the more idiomatic approach.

tasks/swarm_token.rb
chelnak: In this case you have changed from concatenating a string to an array so we should generally lean towards the more idiomatic approach.

tasks/node_ls.rb
```diff
@@ -6,9 +6,9 @@
 require 'puppet'
 
 def node_ls(filter, quiet)
-  cmd_string = 'docker node ls'
-  cmd_string += " --filter=#{filter}" unless filter.nil?
-  cmd_string += ' --quiet' unless quiet.nil?
+  cmd_string = ['docker', 'node', 'ls']
+  cmd_string += [" --filter=#{filter}"] unless filter.nil?
+  cmd_string += [' --quiet'] unless quiet.nil?
```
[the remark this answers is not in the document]
LukasAud: Ah that makes sense

spec/shared_examples/services.rb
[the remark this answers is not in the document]
LukasAud: Yeah, I thought it was strange when I saw it the first time. I just assumed there was a special reason for this set-up.

manifests/run.pp
LukasAud: Placing a space instead of a = works with no errors. Will try to implement late proper segmentation.

manifests/plugin.pp
david22swan: It should be fine so long as the command is sanitized in some other way.
In this case the only variable part, `docker_command` should be given a shell escape as while it has a default value of `docker` it is able to be manually set.
It's a difficult one as it is explicitly meant to be a variable for overriding what the command does....

david22swan: ..... Could add a regex check to it to make sure that the command always includes docker at the end with no white spaces in front of it, i.e. that it's always a path to the docker command

david22swan (approved):
LGTM

PR #871, opened 2022-11-21
(CONT-263) Bumping required puppet version

LukasAud:
Prior to this commit, as part of the codebase hardening project, an important piece of update was missed: raising the minimum required puppet version to 6.24.0, so that certain command breakdowns could occur.

This commit aims to fix that. It can also be considered part of the ticket CONT-263 and the final update required before a major release can be cut.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -59,7 +59,7 @@
   "requirements": [
     {
       "name": "puppet",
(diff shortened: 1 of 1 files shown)


jordanbreen28 (approved):
LGTM. Test failures are present in nightlies

Issue #881, opened 2022-12-13
*_docker_params_changed is not handling/parsing paths/names/destinations correctly


bmagistro:
The *_docker_params_changed is triggering on sequential puppet runs when nothing should have changed resulting in a container restarting every puppet run.

I believe this is very much related to #781 but didn't want to piggy back as that one seems to be discussing issues related to ports/networking and in our case I have narrowed this one down to a different aspect.

## Environment
CentOS 7.9
Puppet 6.28.0
 - docker mod 6.0.2
docker 20.10.17

----

Working backwards on the list of detected changes.  will also preface with I don't do much with ruby so approaches to fix/descriptions may not be ideal.

destinations (https://github.com/puppetlabs/puppetlabs-docker/blob/main/lib/puppet/functions/docker_params_changed.rb#L119) is failing due to the destination map built (https://github.com/puppetlabs/puppetlabs-docker/blob/main/lib/puppet/functions/docker_params_changed.rb#L113) using the wrong index.  In our data (a hash) index 3 maps to the key `Mode` while index 2 maps to the key `Destination`.  To allow position variability it seems like this needs to do the lookup by key instead of by index.

names (https://github.com/puppetlabs/puppetlabs-docker/blob/main/lib/puppet/functions/docker_params_changed.rb#L119) is failing due to a strict order in the comparison of the arrays. This is most easily addressed by adding a `.sort` on the comparison.  It seems like this should be able to be addressed before the comparison when the revised `names` is defined (https://github.com/puppetlabs/puppetlabs-docker/blob/main/lib/puppet/functions/docker_params_changed.rb#L112) but not sure how yet.

paths (https://github.com/puppetlabs/puppetlabs-docker/blob/main/lib/puppet/functions/docker_params_changed.rb#L108) is failing because it is comparing an empty array (`pp_paths`) to a non-empty array (`inspect_paths`).  I haven't dug into this one enough yet to provide a better root cause.

Issue #781, opened 2021-10-22
*_docker_params_changed is notified every time puppet runs, regardless of status


simmerz:
## Describe the Bug
`docker::run` instances are restarted on each puppet run, regardless of whether any params have actually changed. This is using `restart => 'unless-stopped'`, so not using the systemd approach.

## Expected Behavior
Instances should not restart if params have not changed.

## Environment
 - Version 4.1.2
 - Platform Ubuntu 20.04

simmerz:
Is #629 related to this at all?

simmerz:
Hi, I still can't upgrade due to this as docker restarts managed instances every half hour. Am I the only one experiencing this issue?

I'm not using systemd, but using the built in docker restart value

linuxmail:
That is one of the reason, why I switched the whole Docker part to Ansible playbooks. Docker and Puppet isn't fun.

raoulbhatia:
I am having the same issue as described with `puppetserver 6.20.0-1focal` and `puppet-agent 6.28.0-1bionic`
(both from https://apt.puppet.com)

MasterMind2k:
I've been looking around. Change detection for exposed ports is completely incorrect.

The code checks only for `PortBinding` keys, ignoring `HostIP` and `HostPort` part of port bindings defined in puppet.

For instance, code:
```
        ports = inspect_hash['HostConfig']['PortBindings'].keys
        ports = ports.map { |item| item.split('/')[0] }
        pp_ports = opts['ports'].sort if opts['ports'].is_a?(Array)
        pp_ports = [opts['ports']] if opts['ports'].is_a?(String)

        param_changed = true if pp_ports && pp_ports != ports
```

For my case, I have something like this defined in hiera:
```
  ports:
  - <ipv4>:80:5601/tcp
  - "[<ipv6>]:80:5601/tcp"
```

Which results into in docker:
```
{
  "5601/tcp": [
    {
      "HostIp": "<ipv4>",
      "HostPort": "80"
    },
    {
      "HostIp": "<ipv6>",
      "HostPort": "80"
    }
  ]
}
```

If we go through change detection code, it compares the following lists:
```
["<ipv4>:80:5601/tcp", "[<ipv6>]:80:5601/tcp"]
== ["5601"]
```

Which is of course incorrect. Seeing this code, I believe the logic in detection is completely wrong also for other parameters.

MasterMind2k:
[...]

PR #883, opened 2022-12-14
Improve param_changed detection (#881)

bmagistro:
This attempts to fix the issues identified in #881 and while it does address the identified issues in our scenario, I have some concerns about the approach/sanity of these checks related to the concerns called out by @mastermind2k in https://github.com/puppetlabs/puppetlabs-docker/issues/781#issuecomment-1242956951 .

In this case, the comparison for the mounts + volumes does not preserve the source/destination paring so while comparing the list of source and destination mounts might be correct, the pairing could still be wrong ie `source1:dest2` when it should be `source1:dest1` and wound not trigger the `param_changed` event that would be desired.  When talking about configured volumes, It also does not take into account the mount type such as volume, bind, etc (not sure if this one should as it could be multiple types and still be a valid match in my opinion).

It also seems like this function may break when called on a windows host.  All of the start, stop, create, etc calls wrap the command with `run_with_powershell` while the `detect_changes` does not.

I would appreciate hearing from @adrianiurca who authored most of this function as I had to guess at a couple things so if additional light can be shed on what it should be accomplishing I can try to make some updates.

### DIFF
--- a/lib/puppet/functions/docker_params_changed.rb
+++ b/lib/puppet/functions/docker_params_changed.rb
@@ -169,13 +169,58 @@ Puppet::Functions.create_function(:docker_params_changed) do
 
 
         # check if something on ports was changed(some ports were added or removed)
+        # for the ports from inspect, transform this into a hash so we can compare as a set
+        # with one entry per address family
+        ports = []
+        inspect_hash['NetworkSettings']['Ports'].each do |key, value|
+            next if not value
+            cp = key.split('/')
+            value.each do |host_info|
+                entry = {
+                    'protocol' => cp[1],
+                    'host_ip' => host_info.fetch('HostIp'),
+                    'host_port' => host_info.fetch('HostPort'),
+                    'container_port' => cp[0],
+                }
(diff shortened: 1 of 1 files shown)


LukasAud:
Hi @bmagistro, thanks for taking the time to work on this issue. Unfortunately, @adrianiurca is not working in Puppet anymore so we don't think it's going to be possible to get more insight around this update than what we already have.

jst-cyr:
Hey there @bmagistro , during review of old PRs I found this one that didn't get reviewed and I'm wondering if there is still a need or if we can close it?

bmagistro:
I dont remember, we are running a locally patched version and are behind on updating so if it was fixed via other changes would need to re-validate and possibly rework the patch

jst-cyr:
@bmagistro : Thank you for the update! I recognize that could be a pain, but it sounds like the fix is still needed based on your message. Doing a quick check against main looks like it hasn't been solved there yet. If you can resolve conflicts when you have a moment, I can try to get some acceptance testing running on this.

PR #897, opened 2023-02-14
Fix `docker` fact with recent version of docker

smortex:
The `docker` nested fact is supposed to be conditionaly build if the
docker client version is 1.13 or greater, as changed in #166.

Unfortunately, the regexp is not working as expected, and future tweaks
of the regexp did not fix it.

Fix the regexp to match versions greater are equal to 1.13.*.  Add a few
unit tests to verify the outcome.

Fixes #896

PR #900, opened 2023-02-16
Fix version number validation for fact :docker

jaketus:
Docker client versions have reached >=20, and previous validation broke after 19.

I'm not sure why there's the [0-2] at the end, it would imply there has been three digit versions starting with 1 and ending in 0-2, which doesn't really make sense, but I'm too scared of regressions.

I'd guess the ideal would be '[1-9][0-9][\\.\d]+' or simply '[\d\\.]+'


jaketus:
Could be a duplicate of
https://github.com/puppetlabs/puppetlabs-docker/pull/897

PR #921, opened 2023-05-30
pdksync - (MAINT) - Require Stdlib 9.x

LukasAud:
(MAINT) - Update Stdlib dependency
pdk version: `2.7.0`

### DIFF
--- a/lib/puppet/parser/functions/docker_run_flags.rb
+++ b/lib/puppet/parser/functions/docker_run_flags.rb
@@ -10,7 +10,7 @@ module Puppet::Parser::Functions
     escape_function = if self['facts'] && self['facts']['os']['family'] == 'windows'
                         'stdlib::powershell_escape'
                       else
-                        'shell_escape'
+                        'stdlib::shell_escape'
                       end
 
     call_function(escape_function, subject)


LukasAud:
PR currently blocked by main CI failing. Issues are not related to change afaics

bastelfreak:
I think the PR title should be adjusted. this now doesn't allow stdlib 9, it requires it (which is in my opinion a breaking change).

lib/puppet/parser/functions/docker_run_flags.rb
```diff
@@ -8,7 +8,7 @@ module Puppet::Parser::Functions
     subject = args[0]
 
     escape_function = if self['facts'] && self['facts']['os']['family'] == 'windows'
-                        'powershell_escape'
+                        'stdlib::powershell_escape'
                       else
                         'shell_escape'
```
smortex, line 13:
```suggestion
                        'stdlib::shell_escape'
```

smortex (changes requested):
Looks like some namespaced function is missing.  No idea what is going wrong with some versions of windows at this point.

lib/puppet/parser/functions/docker_run_flags.rb
LukasAud: thats odd, I could sweat I searched for all instances of shell_escape previously

This one must have skipped my radar. Nice spot

Ramesh7 (approved):

smortex:
These windows failure seems quite reproducible.  I have basically no experience with docker and windows, but it looks like the command we run do not match the ones from upstream documentation:

https://github.com/puppetlabs/puppetlabs-docker/blob/420e082c870e6eb8eeb62f9f68c13bc527df890c/templates/windows/install_powershell_provider.ps1.erb#L16-L20

vs.

https://github.com/OneGet/MicrosoftDockerProvider/blob/a70c037cd153b8277e2f2d85224d05a044225ad4/README.md?plain=1#L6-L10

Can't say if we do this by mistake or on purpose.

LukasAud:
This PR is failing due to problems present on main. This does not create any new failures on top on the already existing ones afaik. We should be able to merge this and address failures on main in a different place.

Issue #748, opened 2021-05-12
docker::run memory_limit sets a seemingly invalid default value


ALTinners:
Docker's documentation for the `-m` command states

> The maximum amount of memory the container can use. If you set this option, the minimum allowed value is 4m (4 megabyte).

https://docs.docker.com/config/containers/resource_constraints/

However `docker::run` defaults this (optional) argument to a defined value of `0b`

https://github.com/puppetlabs/puppetlabs-docker/blob/b8b0b85213e088137196e5304cac0b4f8495808d/manifests/run.pp#L197

Is there a reasoning as to why this value is set to an apparently invalid number vs leaving it as `undef`? Especially given its already optional

As far as I can tell - Docker is simply ignoring this specified constraint (I have plenty of containers that happily eat several MB of memory) - however that could change and many containers deployed using this module could unexpectedly break.

Vincevrp:
I'm using podman-docker (#562) and this default value prevents containers from starting. Podman is unable to start the container and reports:  `Error: OCI runtime error: unable to start container fcd02362087d56b8829dbf4050c8106985733d06c690720ccbfbe3754323eb9e: sync socket closed`.

The kernel (obviously) complains about `Memory cgroup out of memory`

kenyon:
reopen

PR #927, opened 2023-06-23
Defaults memory_limit to undef instead of ignored 0b value

ALTinners:
## Summary
From https://github.com/puppetlabs/puppetlabs-docker/issues/748 - sets the default `memory_limits` value to `undef` instead of `0b`. As the Docker documentation states that `0b` is an impossible value to use - it appears to be ignoring the `0b` value. In which case this should not be a breaking change to any setup. 

## Additional Context
- Also helps with `podman` usage as per the linked issue

## Related Issues (if any)
https://github.com/puppetlabs/puppetlabs-docker/issues/748 

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -196,7 +196,7 @@ define docker::run (
   Enum[present,absent]                      $ensure                            = 'present',
   Optional[String]                          $verify_digest                     = undef,
   Optional[String]                          $command                           = undef,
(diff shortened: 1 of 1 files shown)


ALTinners:
Realised that the typing of this key has changed from when I last forked it circa 2021 (and implemented this change) - this key is now optional and defaulting to the illegal 0b value. 

This will require a bit more change than a 1 liner. I will reopen when I have that implemented and tested

PR #953, opened 2023-09-26
(maint) Add to supported platform list

barriserloth:
This commit adds to the list of officially supported platform. My rationale for adding Centos/RHEL 8 is because of this change: https://github.com/puppetlabs/puppetlabs-docker/pull/926 My rationale for adding SLES is because of this change: https://github.com/puppetlabs/puppetlabs-docker/pull/591

And I've also added Alma, Oracle, Rocky and Scientific to the list as well since they are all RedHat derivatives and would ideally be supported alongside RHEL.

## Summary
Provide a detailed description of all the changes present in this pull request.
See commit message above

Please let me know if there are any issues with this, and I'm happy to amend my changes. Thanks!

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.


golflimaechoecho:
Is it possible that RHEL8 was excluded from earlier addition as [Docker is not the supported container runtime for RHEL 8 & 9](https://access.redhat.com/solutions/3696691), plus documentation for Docker engine https://docs.docker.com/engine/install/ lists no support for RHEL apart from RHEL (s390x) which is not a supported puppet agent platform

kenyon:
This section of the readme should just point the reader to metadata.json, since that's what actually matters.

jst-cyr:
I updated the branch to resolve conflicts with the main, but I agree with the others that we should be treating the metadata as the source of truth. Currently, the README provides some additional flavour around the limited support, but it doesn't match with what is stated in the metadata. There are additional OS listed as supported not added in the metadata (even in the current main) so I don't think we should take this PR as is.

If we want to add some additional OS, we should update both metadata and the README and make sure they match.

PR #944, opened 2023-08-07
(CAT-1143)-Conversion_of_erb_templates_to_epp

praj1001:
## Summary
Conversion of ERB To EPP for templates in docker module.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified. (For example `puppet apply`)

PR #966, opened 2024-01-08
Avoid spacing changes after switching to EPP

smortex:
In #944, we switched from ERB to EPP.  This has not been released yet,
but while testing I spotted some white space changes that where
unexpected.

Adjust the code so that the EPP template produce the same result as the
ERB template did, making changes more relevant to end-users if they see
some files being changed.

### DIFF
--- a/templates/etc/systemd/system/docker.service.d/service-overrides-debian.conf.epp
+++ b/templates/etc/systemd/system/docker.service.d/service-overrides-debian.conf.epp
@@ -1,8 +1,8 @@
 <% if $service_after_override { -%>
 [Unit]
 After=<%= $service_after_override %>  
-<% } -%>
 
+<% } -%>
 [Service]
 EnvironmentFile=-/etc/default/docker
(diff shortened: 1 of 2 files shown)


kenyon (approved):

jst-cyr:
@smortex  : My apologies that nobody has responded on this for so much time. I'm going through our pull requests across modules and came across this request here. Is this still an issue for you?

smortex:
> Is this still an issue for you?

Not really: 10.0.0 was released with the issue, then @kenyon opened #983 to fix the issue it caused and it was released in 10.0.1 ([and made some people angry because it caused some service restart :shrug:](https://github.com/puppetlabs/puppetlabs-docker/pull/983#issuecomment-2227827645)).

So we can close this one.

kenyon:
Yes, I tried in #978 to get the whitespace restore into the 10.0.0 release.

jst-cyr:
Thanks all! It's good to clean up some of these so I can make sure the team focuses on the right ones. Cheers!

Issue #903, opened 2023-02-27
docker-ce and docker-ce-cli version mismatch when the 'version' option set


vitargelo:
## Describe the Bug
When the version option is set to a specific version, docker-ce-cli package latest version installed.

```
  class { 'docker':
    version        => '5:20.10.21~3-0~ubuntu-focal',
    ip_forward     => false,
    iptables       => false,
    ip_masq        => false,
    log_driver     => 'json-file',
  }
```
And the result:

```
$ docker version

Client: Docker Engine - Community
 Version:           23.0.1
 API version:       1.41 (downgraded from 1.42)
 Go version:        go1.19.5
 Git commit:        a5ee5b1
 Built:             Thu Feb  9 19:46:56 2023
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          20.10.21
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.18.7
  Git commit:       3056208
  Built:            Tue Oct 25 18:00:04 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.18
  GitCommit:        2456e983eb9e37e47538f59ea18f2043c9a73640
 runc:
  Version:          1.1.4
  GitCommit:        v1.1.4-0-g5fd4c4d
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

```

## Expected Behavior
Both, the docker engine and docker CLI have the same defined version (20.10.21)

## Environment
 - Version [6.0.1]
 - Platform [Ubuntu 20.04]

## Additional Context
When trying to install a specific version of the docker-ce package with apt, for some reason it installs the latest docker-ce-cli version as a dependency. Official docker documentation says that we need to set a specific version for both packages in a single command.
```
 VERSION_STRING=5:20.10.13~3-0~ubuntu-jammy
 sudo apt-get install docker-ce=$VERSION_STRING docker-ce-cli=$VERSION_STRING containerd.io docker-buildx-plugin docker-compose-plugin
```
Maybe possible solution is to set-up docker-ce-cli before docker-ce with puppet package resource.

vitargelo:
Related issues: #895 #865

Issue #865, opened 2022-10-17
Should also manage docker-ce-cli package with docker-ce


antaflos:
## Describe the Bug

For Docker CE packages: when managing the Docker package (`docker::manage_package => true`) but *not* setting `docker::package_source` (i.e. packages are installed from APT or YUM repos available to the system) then the `docker-ce-cli` package remains unmanaged by Puppet.

This leads to problems when updating the Docker CE packages (i.e. bumping `docker::version`) down the line, because the `docker-ce-cli` package remains at the version at which it was originally installed. Thus `docker-ce` and `docker-ce-cli` versions drift apart.

This could also be interpreted as a bug in the Docker packages since `docker-ce` doesn't enforce `docker-ce-cli` to be installed in the same version (it just depends on *any* version of `docker-ce-cli`). Maybe this is intentional?

## Expected Behavior

When managing the Docker CE package `docker-ce` without setting `docker::package_source` then the CLI package `docker-ce-cli` should also be managed and set to the same version as the `docker-ce` package.

## Steps to Reproduce

1. Enable Docker APT repo
2. Install Docker CE by means of puppetlabs-docker
3. Bump `docker::version`
4. Observe that `docker-ce-cli` remains at the originally installed version while `docker-ce` is bumped to the version specified in `docker::version`

## Environment
 - Docker 20.10.x
 - puppetlabs-docker 5.0.0
 - Ubuntu 20.04, probably also RHEL and others

## Additional Context

PRs #740 and #827 deal with this issue but only when `docker::package_source` is used. The case when packages are installed from system-wide enabled APT (or YUM) repos is left out.

PR #970, opened 2024-03-04
Ensure docker-ce-cli is same version as docker

viliusgec:
## Summary
When `docker::package_source` is not provided - ensure not only docker version, but also `docker-ce-cli`.

## Additional Context
If `docker::package_source` is not used, then only docker version is ensured and `docker-ce-cli` is left out. That creates issues where docker version gets updated to newer, but `docker-ce-cli` is still using older versions. 

## Related Issues (if any)
[[#865] Should also manage docker-ce-cli package with docker-ce](https://github.com/puppetlabs/puppetlabs-docker/issues/865)
[[#903] docker-ce and docker-ce-cli version mismatch when the 'version' option set ](https://github.com/puppetlabs/puppetlabs-docker/issues/903)

## Checklist

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -80,6 +80,10 @@ class docker::install (
               ensure => $ensure,
               name   => $docker::docker_package_name,
         }))
+        ensure_resource('package', 'docker-ce-cli', stdlib::merge($docker_hash, {
+              ensure => $ensure,
+              name   => $docker::docker_ce_cli_package_name,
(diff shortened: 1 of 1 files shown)


jst-cyr:
Hey there @viliusgec , sorry that it's taken so long for somebody to respond on this PR. I've been reviewing old PRs and was wondering if this was still an issue for you? Is this fix still needed?

PR #902, opened 2023-02-24
Update docker compose

davidphay:
Update docker compose to the latest version (2.16.0) & install it as docker plugin on linux server since compose v1 will be deprecated on June 2023.

reference: https://docs.docker.com/compose/install/linux/

Issue #891, opened 2023-01-30
Support `docker-compose-plugin` as provider for `docker_compose` resource


kenyon:
## Use Case

The `docker_compose` resource in this module should support the Docker Compose plugin.

## Describe the Solution You Would Like

The `docker_compose` resource uses docker-compose-plugin rather than failing with `Error: Could not find a suitable provider for docker_compose`.

## Describe Alternatives You've Considered

Installing standalone Docker Compose would be an alternative, but that is not supported by upstream per https://docs.docker.com/compose/install/ and upstream provides `docker-compose-plugin` as packages per https://docs.docker.com/compose/install/linux/.

## Additional Context

See also:  #839

vegaaz:
This should be focused. According to the `[documentation](https://docs.docker.com/compose/compose-file/) docker-compose v1 is going to loose support after june 2023.

kenyon:
@vegaaz since v1 is unsupported upstream, removing support in this module would be 👍 from me.

robertc99:
docker-ce ships with a version of docker-compise on /usr/libexec/docker/cli-plugins/docker-compose
but the module doesnt use it by default since its not on the standard path.
If you change your path to include /usr/libexec/docker/cli-plugins/ before running puppet then docker_compose will find it and use it. 

So all that needs to be done is adjust the default path used by the docker_compose resource
But Im not sure how to do that.

kenyon:
`/usr/libexec` is not supposed to be in `PATH`. I think it's best to stick with installing `docker-compose-plugin` per upstream documentation.

kenyon:
Also, at least with Ubuntu and Red Hat packages provided by Docker, `/usr/libexec/docker/cli-plugins/docker-compose` is provided by the `docker-compose-plugin` package, not the `docker-ce` package.

robertc99:
Ok, the docker-compose is installed by the docker-compose-plugin package, not docker-ce itself.
But even though it is installed, the docker module won't use it. 

How do we get the module to use it if we dont change the path.
Should the module create a link to it in /usr/local/bin?

robertc99:
I checked the upstream documentation, which recommends installing the plugin package. But even when you do that, the module doesnt use it.

kenyon:
> I checked the upstream documentation, which recommends installing the plugin package. But even when you do that, the module doesnt use it.

That's the point of this issue. The module needs to be updated so that it uses `docker compose` (with a space).

[...]

PR #975, opened 2024-04-11
Use the docker-compose-plugin via 'docker compose' instead of 'docker-compose'

nathanlcarlson:
## Summary

Use the docker-compose-plugin via 'docker compose' instead of 'docker-compose'

From https://docs.docker.com/compose/install/
> Docker's documentation refers to and describes Compose V2 functionality. Effective July 2023, Compose V1 stopped receiving updates and is no longer in new Docker Desktop releases. Compose V2 has replaced it and is now integrated into all current Docker Desktop versions. For more information, see [Migrate to Compose V2](https://docs.docker.com/compose/migrate).

## Additional Context
Based on #902. Created my own PR to hopefully get this resolved sooner.

## Related Issues (if any)
#891 

## Checklist
- [x] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/compose.pp
+++ b/manifests/compose.pp
@@ -40,7 +40,7 @@ class docker::compose (
         }
       }
       'Windows': {
-        fail('Docker compose is installed with docker machine on Windows')
+        fail('Docker compose is installed with Docker Desktop on Windows')
       }
       default: {
         fail('This module only works on Debian, RedHat or Windows.')
--- a/manifests/compose.pp
+++ b/manifests/compose.pp
@@ -20,21 +20,23 @@ class docker::compose (
 
     case $facts['os']['family'] {
       'Debian': {
+        $_require = $docker::use_upstream_package_source ? {
+          true  => Apt::Source['docker'],
+          false => undef,
+        }
         package { 'docker-compose-plugin':
           ensure  => $package_ensure,
-          require => $docker::use_upstream_package_source ? {
-            true  => Apt::Source['docker'],
-            false => undef,
-          },
+          require => $_require,
         }
       }
       'RedHat': {
+        $_require = $docker::use_upstream_package_source ? {
+          true  => Yumrepo['docker'],
+          false => undef,
+        }
         package { 'docker-compose-plugin':
           ensure  => $package_ensure,
-          require => $docker::use_upstream_package_source ? {
-            true  => Yumrepo['docker'],
-            false => undef,
-          },
+          require => $_require,
         }
       }
       'Windows': {
--- a/manifests/compose.pp
+++ b/manifests/compose.pp
@@ -11,6 +11,8 @@ class docker::compose (
   Enum[present,absent] $ensure  = present,
   Optional[String]     $version = undef,
 ) {
+  include docker
+
   if $docker::manage_package {
     include docker::params
 
--- a/manifests/compose.pp
+++ b/manifests/compose.pp
@@ -12,6 +12,8 @@ class docker::compose (
   Optional[String]     $version = undef,
 ) {
   if $docker::manage_package {
+    include docker::params
+
     $_version = $version ? {
       undef   => $docker::params::compose_version,
       default => $version,
--- a/manifests/compose.pp
+++ b/manifests/compose.pp
@@ -22,7 +22,7 @@ class docker::compose (
       'Debian': {
         package { 'docker-compose-plugin':
           ensure  => $package_ensure,
-          require => defined(bool2str($docker::use_upstream_package_source)) ? {
+          require => $docker::use_upstream_package_source ? {
             true  => Apt::Source['docker'],
             false => undef,
           },
@@ -31,7 +31,7 @@ class docker::compose (
       'RedHat': {
         package { 'docker-compose-plugin':
           ensure  => $package_ensure,
-          require => defined(bool2str($docker::use_upstream_package_source)) ? {
+          require => $docker::use_upstream_package_source ? {
             true  => Yumrepo['docker'],
             false => undef,
           },
--- a/manifests/compose.pp
+++ b/manifests/compose.pp
@@ -7,118 +7,42 @@
 # @param version
 #   The version of Docker Compose to install.
 #
-# @param install_path
-#   The path where to install Docker Compose.
-#
-# @param symlink_name
-#   The name of the symlink created pointing to the actual docker-compose binary
-#   This allows use of own docker-compose wrapper scripts for the times it's
-#   necessary to set certain things before running the docker-compose binary
-#
-# @param proxy
-#   Proxy to use for downloading Docker Compose.
-#
-# @param base_url
-#   The base url for installation
-#   This allows use of a mirror that follows the same layout as the
-#   official repository
-#
-# @param raw_url
-#   Override the raw URL for installation
-#   The default is to build a URL from baseurl. If rawurl is set, the caller is
-#   responsible for ensuring the URL points to the correct version and
-#   architecture.
-#
-# @param curl_ensure
-#   Whether or not the curl package is ensured by this module.
-#
 class docker::compose (
-  Enum[present,absent] $ensure                 = present,
-  Optional[String]               $version      = $docker::params::compose_version,
-  Optional[String]               $install_path = $docker::params::compose_install_path,
-  Optional[String]               $symlink_name = $docker::params::compose_symlink_name,
-  Optional[Pattern['^((http[s]?)?:\/\/)?([^:^@]+:[^:^@]+@|)([\da-z\.-]+)\.([\da-z\.]{2,6})(:[\d])?([\/\w \.-]*)*\/?$']] $proxy = undef,
-  Optional[String]               $base_url     = $docker::params::compose_base_url,
-  Optional[String]               $raw_url      = undef,
-  Optional[Boolean]              $curl_ensure  = $docker::params::curl_ensure,
+  Enum[present,absent] $ensure  = present,
+  Optional[String]     $version = $docker::params::compose_version,
 ) inherits docker::params {
-  if $facts['os']['family'] == 'windows' {
-    $file_extension = '.exe'
-    $file_owner     = 'Administrator'
-  } else {
-    $file_extension = ''
-    $file_owner     = 'root'
-  }
-
-  $docker_compose_location           = "${install_path}/${symlink_name}${file_extension}"
-  $docker_compose_location_versioned = "${install_path}/docker-compose-${version}${file_extension}"
-
-  if $ensure == 'present' {
-    if $raw_url != undef {
-      $docker_compose_url = $raw_url
+  if $docker::manage_package {
+    if $version and $ensure != 'absent' {
+      $package_ensure = $version
     } else {
-      $docker_compose_url = "${base_url}/${version}/docker-compose-${facts['kernel']}-${facts['os']['hardware']}${file_extension}"
+      $package_ensure = $ensure
     }
 
-    if $proxy != undef {
-      $proxy_opt = "--proxy ${proxy}"
-    } else {
-      $proxy_opt = ''
-    }
-
-    if $facts['os']['family'] == 'windows' {
-      $docker_download_command = "if (Invoke-WebRequest ${docker_compose_url} ${proxy_opt} -UseBasicParsing -OutFile \"${docker_compose_location_versioned}\") { exit 0 } else { exit 1}" # lint:ignore:140chars
-
-      $parameters = {
-        'proxy'                             => $proxy,
-        'docker_compose_url'                => $docker_compose_url,
-        'docker_compose_location_versioned' => $docker_compose_location_versioned,
+    case $facts['os']['family'] {
+      'Debian': {
+        package { 'docker-compose-plugin':
+          ensure  => $package_ensure,
+          require => defined(bool2str($docker::use_upstream_package_source)) ? {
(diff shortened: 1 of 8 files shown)


nathanlcarlson:
This link is dead https://puppetlabs.github.io/iac/docs/contributing_to_a_module.html

nathanlcarlson:
I've done some basic manual testing of this now, both with a fresh start of a compose project and with a running compose project.

.rubocop.yml
```diff
@@ -529,6 +529,11 @@ Lint/DuplicateBranch:
   Enabled: false
 Lint/DuplicateMagicComment:
   Enabled: false
+<<<<<<< HEAD
+Lint/DuplicateMatchPattern:
+  Enabled: false
+=======
+>>>>>>> 8f4b3d98c7110e3a4fc9151a14e55876eec2b585
```
smortex, line 536: Merge conflicts?

smortex (changes requested):
It's better to focus on a single point for each PR (to minimize the changes to review) with commits that change a single aspect at a time (to make review easier).

For example 91ab09c2e9f40cea3351b150afe07e25bb7c289d does multiple things:
* It updates PDK-related files (IMHO this should not be part of this PR unless if it fix CI failure)
* It update lib/puppet/provider/docker_compose/ruby.rb to remove [unused code](https://github.com/puppetlabs/puppetlabs-docker/pull/975/commits/91ab09c2e9f40cea3351b150afe07e25bb7c289d#diff-0b884d248175f3ad6cf5342c74d8d766c5b6ab28ffdb31f369c614b67b5f36ddL12) (which should be part of this PR).

Do you think you can rework this PR and adjust the commits you add on top of #975 to only finish the work that was started by @davidphay, ignoring the PDK stuff, and without the merge commit?  Maybe it is enough to have the CI green :wink: 

Thanks!

nathanlcarlson:
Sure I can do those things. Please fix your CONTRIBUTING.md.

nathanlcarlson:
I've made a proper mess by reverting commits that I'd merged. I'll figure out a better way.

manifests/compose.pp
```diff
@@ -7,118 +7,30 @@
...
-    if $facts['os']['family'] == 'windows' {
-      $docker_download_command = "if (Invoke-WebRequest ${docker_compose_url} ${proxy_opt} -UseBasicParsing -OutFile \"${docker_compose_location_versioned}\") { exit 0 } else { exit 1}" # lint:ignore:140chars
-
-      $parameters = {
-        'proxy'                             => $proxy,
-        'docker_compose_url'                => $docker_compose_url,
-        'docker_compose_location_versioned' => $docker_compose_location_versioned,
+    case $facts['os']['family'] {
+      'Debian': {
+        ensure_packages('docker-compose-plugin', { ensure => $package_ensure, require => defined(bool2str($docker::use_upstream_package_source)) ? { true => Apt::Source['docker'], false => undef } }) #lint:ignore:140chars
```
kenyon, line 23: We shouldn't be using `ensure_packages` in this module. It should be expected that this module manages the package resources.

kenyon (changes requested):
A rebase on main will get rid of the merge commit.

There are changes in early commits that are then removed in later commits. It would be good to clean up the commit history.

manifests/compose.pp
nathanlcarlson: I've modified it to use ensure_resource as I found in install.pp.

nathanlcarlson:
Aren't changes to previous commits the nature of applying requested changes? Besides remaking the branch and PR altogether I'd have to really stretch my git abilities to make it appear such that it was all done as desired from the beginning.

nathanlcarlson:
I suppose maybe you want me to squash the commits together selectively? I can try that.

nathanlcarlson:
Looking into squashing it, I'm not sure that will comply with "commits that change a single aspect at a time". I won't do anything without further advisement.

manifests/compose.pp
kenyon: I meant that the `package` resource should be used, not `ensure_resource` nor `ensure_packages`.

kenyon:
Use interactive rebase to edit the commit history. I'm not saying squash all of the changes into one commit, unless that is what makes the most sense.

manifests/compose.pp
nathanlcarlson: Okay. I will do it differently than the way it is done in manifests/install.pp.

nathanlcarlson:
Unfortunately, it really only made sense to me to make it into a single commit.

manifests/compose.pp
```diff
@@ -7,118 +7,42 @@
...
-
-      $parameters = {
-        'proxy'                             => $proxy,
-        'docker_compose_url'                => $docker_compose_url,
-        'docker_compose_location_versioned' => $docker_compose_location_versioned,
+    case $facts['os']['family'] {
+      'Debian': {
+        package { 'docker-compose-plugin':
+          ensure  => $package_ensure,
+          require => defined(bool2str($docker::use_upstream_package_source)) ? {
```
kenyon, line 25: Isn't this equivalent?

```suggestion
          require => $docker::use_upstream_package_source ? {
```

manifests/compose.pp
```diff
@@ -7,118 +7,42 @@
...
-      exec { "Install Docker Compose ${version}":
-        path    => '/usr/bin/',
-        cwd     => '/tmp',
-        command => "curl -s -S -L ${proxy_opt} ${docker_compose_url} -o ${docker_compose_location_versioned}",
-        creates => $docker_compose_location_versioned,
-        require => Package['curl'],
+      'RedHat': {
+        package { 'docker-compose-plugin':
+          ensure  => $package_ensure,
+          require => defined(bool2str($docker::use_upstream_package_source)) ? {
```
kenyon, line 34:
```suggestion
          require => $docker::use_upstream_package_source ? {
```

kenyon (commented):
The `be_include` to `include` changes should be a separate commit, since they are unrelated to the other changes.

A couple of comments, otherwise looks good.

manifests/compose.pp
nathanlcarlson: I thought that was a bit overly complicated as well. I've changed it.

nathanlcarlson:
I removed the `be_include` changes altogether as neither my `pdk validate` and `pdk test unit` complained about it after reverting.

kenyon:
Looks OK, just need to get someone from Perforce to approve the workflow run to see test results.

rwaffen:
Nice work! I'm curious about this change too!

kenyon:
@nathanlcarlson standalone docker-compose has been deprecated for a while (if I'm understanding you correctly), so I don't think there is any upgrading to do for that.

smortex:
> Perhaps upgrading the default installed version of the standalone docker-compose may be a preferable path forward?

IMHO default versions are not great because bumping them is a backwards-incompatible change, so it basically never happens.  When they are installed as packages, the default value of `present` is reasonable because it allows to follow package updates and still allow users to pin a specific version if they have to.  I like it a lot.

> It would be nice to have the compose plugin installed as a package as it is then managed and updated as any other package would be on Linux.

:100:

I understand there are 2 issues here:
1. the installation method is not great; and
2. the default version is not great.

Improving the installation method being backwards-incompatible, changing the version it install by default looks fine to me.

Upgrading to the next major version of the modules for users of docker-compose would require to:
- Manually remove the `docker-compose` program that the older version of the module installed (or maybe we can automate this removal);
- Explicitly indicate which version of the docker-compose-plugin to install if needed.

This feels reasonable to me.

nathanlcarlson:
@kenyon Ahh, yes, I see that the Docker site reports it as "not recommended and is only supported for backward compatibility purposes".

I've applied some changes that attempt to repair and cleanup the acceptance tests.

manifests/compose.pp
```diff
@@ -7,118 +7,50 @@
...
-  if $ensure == 'present' {
-    if $raw_url != undef {
-      $docker_compose_url = $raw_url
-    } else {
-      $docker_compose_url = "${base_url}/${version}/docker-compose-${facts['kernel']}-${facts['os']['hardware']}${file_extension}"
+  Enum[present,absent] $ensure  = present,
+  Optional[String]     $version = undef,
+) {
+  if $docker::manage_package {
+    include docker::params
```
smortex, line 15: `params.pp` is something deprecated, and was used by inheriting from it:

```puppet
class docker::compose (
  Optional[String]     $version = $docker::params::version, # 2. set default value
) inherits docker::params {  # 1. inherit
  warning("version from params.pp with user override is ${version}") # 3. use value
}
```

This is arguably inelegant.  Better skip `params.pp` completly:

```puppet
class docker::compose (
  String[1]     $version = 'installed',
) {
  warning("version is ${version}")
}
```

manifests/compose.pp
```diff
@@ -7,118 +7,50 @@
...
-      $docker_download_command = "if (Invoke-WebRequest ${docker_compose_url} ${proxy_opt} -UseBasicParsing -OutFile \"${docker_compose_location_versioned}\") { exit 0 } else { exit 1}" # lint:ignore:140chars
-
-      $parameters = {
-        'proxy'                             => $proxy,
-        'docker_compose_url'                => $docker_compose_url,
-        'docker_compose_location_versioned' => $docker_compose_location_versioned,
+    case $facts['os']['family'] {
+      'Debian': {
+        $_require = $docker::use_upstream_package_source ? {
+          true  => Apt::Source['docker'],
```
smortex, line 30: If I recall correctly, we must require `Class['apt::update']` otherwise Puppet can try to install the package before `apt update` is run and fail or install the wrong package.

manifests/compose.pp
```diff
@@ -7,118 +7,50 @@
...
+          false => undef,
+        }
+        package { 'docker-compose-plugin':
+          ensure  => $package_ensure,
+          require => $_require,
+        }
```
smortex, line 36: Duplicate code.

manifests/compose.pp
```diff
@@ -7,118 +7,50 @@
...
-        require => Package['curl'],
+      'RedHat': {
+        $_require = $docker::use_upstream_package_source ? {
+          true  => Yumrepo['docker'],
+          false => undef,
+        }
+        package { 'docker-compose-plugin':
+          ensure  => $package_ensure,
+          require => $_require,
+        }
```
smortex, line 46: Duplicate code.  We can have a single copy outside of the case statement.

manifests/compose.pp
```diff
@@ -7,118 +7,50 @@
...
-
-  if $ensure == 'present' {
-    if $raw_url != undef {
-      $docker_compose_url = $raw_url
-    } else {
-      $docker_compose_url = "${base_url}/${version}/docker-compose-${facts['kernel']}-${facts['os']['hardware']}${file_extension}"
+  Enum[present,absent] $ensure  = present,
+  Optional[String]     $version = undef,
+) {
+  if $docker::manage_package {
```
smortex, line 14: We need to `include docker` before if we want to use `$docker::variable`.  That being said, what would be the point of including `docker::compose` if we don't want to install packages?  Can't we just get rid of the condition?

nathanlcarlson: The install.pp manifest looks pretty similar, essentially all wrapped in this `manage_package` condition. I've notice `Apt::Source['docker']` is only declared if `use_upstream_package_source` is true and `include apt` only happens if `manage_package` is true as well. `Yumrepo['docker']` is only declared if both `use_upstream_package_source` and `manage_package` are true.

manifests/compose.pp
nathanlcarlson: It looks like the [apt docs](https://forge.puppet.com/modules/puppetlabs/apt/reference#-apt--source--notify_update) have `notify_update` default to true on the `apt::source` resource.

rwaffen:
is there something that could be done to help to move this forward?

nathanlcarlson:
> is there something that could be done to help to move this forward?

I was waiting for @smortex to reply on my response to their comments. I can clean up duplicate code, but I made a couple points about the other comments.

smortex:
> I was waiting for @smortex to reply on my response to their comments. I can clean up duplicate code, but I made a couple points about the other comments.

Dang! Too many notifications :-)  I don't "own" the code, so feel free to add changes which allow other contributors to provide feedback too, and do not wait for me as my availability is somewhat sparse currently :-D  I will reply at the right location to your points.

manifests/compose.pp
smortex: Yes, this is the other side of the requirements: by default, apt::source changes trigger an apt update, but the package installation can happen at any time, unless we say we want it to be after the apt update.

Similar case in another module: 

https://github.com/voxpupuli/puppet-gitlab/blob/master/manifests/omnibus_package_repository.pp#L58

manifests/compose.pp
smortex: Yeh,  `docker::install` seems to have the same issue.  It should probably `include docker`, or maybe be a private class.

This module is not in a good shape unfortunately

rh-org:
Any news here? Any help needed? Would be great to get this finally resolved.

nathanlcarlson:
I've pushed some commits that attempt to make the requested changes.

Ramesh7 (approved):
The changes looks good to me, but will wait for @smortex @kenyon feedback here.

kenyon (approved):
Did some tests with this code on a node where I use `docker_compose`, worked for me!

PR #977, opened 2024-04-18
(CDPE-6220) Fix variable references in run.pp

eputnam:
## Summary

Currently, CD4PE is using this module as part of the version 5 installation. When we use docker::run, as part of starting up the application containers, we're getting warnings like this in the customer-facing terminal output:

Unknown variable: 'docker::docker_group'. (file: /Users/eric.putnam/ws/cd4pe/PipelinesInfra/deploy/envs/cdpe6633/.modules/docker/manifests/run.pp, line: 268, column: 19)
Unknown variable: 'docker::service_name'. (file: /Users/eric.putnam/ws/cd4pe/PipelinesInfra/deploy/envs/cdpe6633/.modules/docker/manifests/run.pp, line: 267, column: 19)

This doesn't make for a very good customer experience. Since these vars are in docker::params, I added "::params" to the two lines where docker_group and service_name are assigned and that seemed to take care of the warnings. Please correct me if I'm wrong about this change, our main goal is to get rid of these warnings.

## Additional Context
These warnings can be seen by running the following bolt plan, which requires this module:
```
plan whalesay::docker_run() {
  apply_prep('localhost')

  apply('localhost') {
    docker::run{'example run':
      image            => 'docker/whalesay',
      pull_on_start    => true,
      restart          => "no",
    }
  }
}
```

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/run.pp
+++ b/manifests/run.pp
@@ -264,8 +264,8 @@ define docker::run (
     $docker_command = $docker::params::docker_command
   }
 
-  $service_name = $docker::service_name
-  $docker_group = $docker::docker_group
+  $service_name = $docker::params::service_name
+  $docker_group = $docker::params::docker_group
 
   if $restart {
     assert_type(Pattern[/^(no|always|unless-stopped|on-failure)|^on-failure:[\d]+$/], $restart)


manifests/run.pp
```diff
@@ -264,8 +264,8 @@
     $docker_command = $docker::params::docker_command
   }
 
-  $service_name = $docker::service_name
-  $docker_group = $docker::docker_group
+  $service_name = $docker::params::service_name
+  $docker_group = $docker::params::docker_group
```
smortex, line 268: [The `docker` class inherits form `docker::params`](https://github.com/puppetlabs/puppetlabs-docker/blob/8f4b3d98c7110e3a4fc9151a14e55876eec2b585/manifests/init.pp#L476) and [allow personalization of these parameters](https://github.com/puppetlabs/puppetlabs-docker/blob/8f4b3d98c7110e3a4fc9151a14e55876eec2b585/manifests/init.pp#L439-L442) (using the values from `docker::params` for default values if not provided).

In order to access `$docker::*` variables, including this class should be enough:

```puppet
  include docker
```

smortex (changes requested):
Directly using `$docker::params::*` is not correct because it contain a default value for a parameter the end user may have provided a specific one.

I added details and a better approach in in-line comments.

Thanks!

kenyon:
Also, `params.pp` should be eliminated, like what I've done for the APT module in https://github.com/puppetlabs/puppetlabs-apt/pull/1172. This docker module would take a lot more effort though.

jst-cyr:
@eputnam : Do you think this change in the docker module would still be valuable? This PR was unfortunately not looked into for quite some time and I'm wondering if it's still needed?

jst-cyr:
Closing this PR for now until a need rises for this in the future at which point we can re-open!

PR #983, opened 2024-07-03
systemd service overrides: restore whitespace to pre-conversion

kenyon:
When these templates were converted from ERB to EPP in commit 3e101cb3151c3e739cd8541dfbc5329f6ae6b647 (#944), these lines were swapped, causing a whitespace change and unnecessary service restarts.


bastelfreak (approved):

Ramesh7 (approved):

weastur:
Well, now it causing restart while upgrading `10.0.0` -> `10.0.1`. 
It was major upgrade `9` -> `10`, so the restart, despite its useless, was acceptable.
Now, the minor upgrade cause restart.
Please, don't do like this anymore.

Issue #984, opened 2024-07-08
v10 uses docker arguments and options in the wrong order.


augustocarvalho:
## Describe the Bug
v10 breaks puppet during the docker compose check if exists.
The argument array is adding the docker options between the -p and <name>.

The array has now 5 elements and it inserts in a wrong position (3):
args = ['compose', compose_files, '-p', name, 'config'].insert(3, resource[:options]).compact

#Puppet debug:
Info: Checking for compose project portainer
Debug: Executing: '/usr/bin/docker compose -f /var/docker/portainer/docker-compose.yml -p --project-directory=/var/docker/portainer portainer config'
Error: /Stage[main]/Profiles::System::Docker_server/Apps::Docker::Portainer[portainer]/Apps::Docker::Webapp[portainer]/Apps::Docker::App[portainer]/Docker_compose[portainer]: Could not evaluate: (<unknown>): could not find expected ':' while scanning a simple key at line 4 column 1
Wrapped exception:
(<unknown>): could not find expected ':' while scanning a simple key at line 4 column 1

## Expected Behavior
Arguments array should insert the docker options after the -p <name>. (Array position is now 4.)
args = ['compose', compose_files, '-p', name, 'config'].insert(4, resource[:options]).compact

# Puppet debug:
Info: Checking for compose project portainer
Debug: Executing: '/usr/bin/docker compose -f /var/docker/portainer/docker-compose.yml -p portainer --project-directory=/var/docker/portainer config'

## Steps to Reproduce
Declare a docker_compose resource with options (for example --project-directory)

## Environment
 - Version 10.0.0
 - Platform Ubuntu 20.04
 - Puppet7

PR #988, opened 2024-08-07
Move up_args after the up command.

ghost:
## Summary
Fixes issue #984 where up_args are being placed before up, and need to go after.

## Additional Context
Since 'compose' became an argument, the insert statement just need to be incremented by one to represent the new position.  The other insert seems fine where it is.

## Related Issues (if any)
#984

### DIFF
--- a/lib/puppet/provider/docker_compose/ruby.rb
+++ b/lib/puppet/provider/docker_compose/ruby.rb
@@ -74,7 +74,7 @@ Puppet::Type.type(:docker_compose).provide(:ruby) do
 
   def create
     Puppet.info("Running compose project #{name}")
-    args = ['compose', compose_files, '-p', name, 'up', '-d', '--remove-orphans'].insert(3, resource[:options]).insert(5, resource[:up_args]).compact
+    args = ['compose', compose_files, '-p', name, 'up', '-d', '--remove-orphans'].insert(3, resource[:options]).insert(6, resource[:up_args]).compact
     docker(args)
     return unless resource[:scale]
 


lib/puppet/provider/docker_compose/ruby.rb
```diff
@@ -74,7 +74,7 @@ def get_image(service_name, compose_services)
 
   def create
     Puppet.info("Running compose project #{name}")
-    args = ['compose', compose_files, '-p', name, 'up', '-d', '--remove-orphans'].insert(3, resource[:options]).insert(5, resource[:up_args]).compact
+    args = ['compose', compose_files, '-p', name, 'up', '-d', '--remove-orphans'].insert(3, resource[:options]).insert(6, resource[:up_args]).compact
```
smortex, line 77: The usage of `#insert` make it a bit hard to follow.  I think we could benefit from re-writing this as something like:

```ruby
args = ['compose', compose_files, '-p', name, resource[:options], 'up', resource[:up_args], '-d', '--remove-orphans'].flatten.compact
```
(assuming the index of the first insert is wrong because I don't think it is correct to insert `resource[:options]` between `-p` and `name`).

Also consistently using long options helps for readability: `-p` => `--project-name`, `-d` => `--detach`

ncstate-daniel: If y'all want to do it a different way that's certainly cool with me.  =)  I needed this fixed asap and submitted my fix without modifying the way things are done too much.

ncstate-daniel: (fwiw i agree -- i don't entirely understand why it was done the way it was)

jst-cyr:
This PR was never approved or updated by the original contributor and now has conflicts that must be resolved. The original contributor is now a deleted user. Going to close this for now and if it is needed a new PR can be opened.

PR #995, opened 2024-09-05
Allow --icc=false option for docker daemon

### LINKED ISSUE #989: Unpossible to disable icc with 10.x version of this module
## Describe the Bug
When the parameter docker::icc is set to false, this parameter is completely ignored

## Expected Behavior
The OPTIONS variable in /etc/sysconfig/docker should include --icc=false

## Steps to Reproduce
1. docker::icc: false in hiera
2. Upgrade to 10.x version of this module
3. See the --icc=false parameter disapear in /etc/sysconfig/docker

## Environment
 - Version 10.0.0
 - Platform Almalinux 9

## Additional Context
The conversion from erb to epp replaced:
`<% unless @icc == nil %> --icc=<%= @icc %><% end -%>`
with:
`<% if $icc { %> --icc=<%= $icc %><% } -%>`
Which behaves completely different when icc is false. A workaround would be to set docker::icc to the string "false", but that will not work because manifests/init.pp demands it to be a boolean.

bjvrielink:
This used to be possible in version 9.1.0 and older, but got lost in translation from .erb to .epp templates.

Fixes #989

## Summary
Changes 5 templates to use unless instead of if for the --icc parameter

## Additional Context
If a user wants to specify the --icc parameter, they always will want to use --icc=false, as the default is true.
By using `if $icc` in the template, it is impossible to set -icc=false.

## Related Issues (if any)
See issue #989 

## Checklist
- [x] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [X] Manually verified. (For example `puppet apply`)

### DIFF
--- a/templates/etc/conf.d/docker.epp
+++ b/templates/etc/conf.d/docker.epp
@@ -11,7 +11,7 @@ other_args="<% -%>
  --ip-forward=<%= $ip_forward -%>
  --iptables=<%= $iptables -%>
  --ip-masq=<%= $ip_masq -%>
-<% if $icc { %> --icc=<%= $icc %><% } -%>
+<% unless $icc == undef { %> --icc=<%= $icc %><% } -%>
 <% if $fixed_cidr { %> --fixed-cidr <%= $fixed_cidr %><% } -%>
 <% if $default_gateway { %> --default-gateway <%= $default_gateway %><% } -%>
 <% if $bridge { %> --bridge <%= $bridge %><% } -%>
--- a/templates/etc/conf.d/docker.gentoo.epp
+++ b/templates/etc/conf.d/docker.gentoo.epp
@@ -11,7 +11,7 @@ DOCKER_OPTS="<% -%>
  --ip-forward=<%= $ip_forward -%>
  --iptables=<%= $iptables -%>
  --ip-masq=<%= $ip_masq -%>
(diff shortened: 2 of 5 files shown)


jst-cyr:
My apologies, @bjvrielink that it took so long for somebody to reach out about this PR. I'm going through old PRs and discovered this one. I'm wondering if this is still a use case you need solved?

bjvrielink:
yes, please!

jst-cyr:
Test PR created to run CI against this change: https://github.com/puppetlabs/puppetlabs-docker/pull/1069

jst-cyr:
**Acceptance Test Results**: https://github.com/puppetlabs/puppetlabs-docker/actions/runs/33518659925?pr=1069

CENTOS 7, CENTOS 8, Ubuntu 20.04, and Debian 10 are failing in a matching way to failures seen on the nightly CI against main. This PR doesn't seem to be introducing those failures.

Windows 2022, however, is failing in that PR even though it succeeded on the latest nightly runs. The failure doesn't seem related to the PR, given the connectivity issue that is showing in the logs, but I have tried to re-run it a few times and if it's a temporary network flakiness issue it doesn't seem to be resolving.

I'm going to leave that CI PR open for a bit and try rerunning again later to see if the networking issue has resolved, but we may need to dig into what is going on with Windows 2022 here.

PR #994, opened 2024-09-04
Update docker.rb for Podman

adamboutcher:
## Summary

This if statement will enable the fact to work if you are using podman with docker symlink/support as the json syntax is lowercase on podman but camel/title case on docker.

## Related Issues (if any)
https://github.com/puppetlabs/puppetlabs-docker/issues/982
Possibly this too https://github.com/puppetlabs/puppetlabs-docker/issues/749

Provide a detailed description of all the changes present in this pull request.
Just an if statement in the factor code (docker.rb)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [X] Manually verified. (For example `puppet apply`)

PR #1005, opened 2024-11-27
Update docker.rb for Podman

malikparvez:
## Summary
This if statement will enable the fact to work if you are using podman with docker symlink/support as the json syntax is lowercase on podman but camel/title case on docker.

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.


kenyon:
Why are you redoing #994?

malikparvez:
> Why are you redoing #994?

ohh this is just a draft PR as the original was failing I was just testing the pipelines somehow having some trouble with my local system.

PR #1010, opened 2025-01-15
Use networking.interfaces fact

vStone:
## Summary
Remove use of the interfaces fact but rely on networking.interfaces

## Additional Context
See https://www.puppet.com/docs/puppet/8/upgrading-from-puppet7-to-puppet8#upgrading-from-puppet7-to-puppet8-legacy-facts-deprecation

## Checklist
- [x] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/lib/facter/docker.rb
+++ b/lib/facter/docker.rb
@@ -38,7 +38,7 @@ docker_command = if Facter.value(:kernel) == 'windows'
                  end
 
 def interfaces
-  Facter.value(:interfaces).split(',')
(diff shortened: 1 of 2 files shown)


jst-cyr:
Sorry for the long delay in response on this PR. It looks like a good change and I think we should go forward with this. 
@cekstam : Do you think you could provide some details on the issues you are having with this right now? That would help with the prioritization.

PR #1072, opened 2026-09-11
Detect docker_stack presence from swarm services, not docker ps.

### LINKED ISSUE #968: docker_stack incorrectly determines if the stack is deployed
## Describe the Bug
When a stack runs on several machines and the services are spread out between them, this check in `docker/lib/puppet/provider/docker_stack/ruby.rb` doesn't work correctly:

```
      if stack_services.count != stack_containers.count
        return false
      end
```

## Expected Behavior
Containers belonging to the same stack but running on other machines should be taken into account.

## Additional Context
This code exists in version 4 of the module but also in the current version. I'm prepared to learn that I'm using the stack resource incorrectly for some reason.

### LINKED ISSUE #973: docker_stack and docker_compose use outdated way of using psych's YAML.safe_load
## Describe the Bug
This module inherited a bug in PDK which causes both the `docker_stack` and `docker_compose` resources to ungracefully fail with a "wrong number of arguments" error while calling `psych`'s YAML.safe_load.

See the following issues:

https://github.com/puppetlabs/pdk/issues/1143
https://github.com/ruby/psych/discussions/571

It looks like this issue was fixed in 2021, however, this module is still using the old way of calling safe_load in both of these resources. Any recent installation of psych (which is a dependency of the PDK) will error out.

## Expected Behavior
I should be able to use docker_stack or docker_compose without an error.

## Steps to Reproduce
Deploy a docker_stack or docker_compose resource with a version of psych greater than 4 on the host system. You can see the version of the gem installed in `/opt/puppetlabs/puppet/lib/ruby/gems/3.2.0/gems` although your installed version of ruby may be different.

## Environment
 - Version 9.1
 - Platform Ubuntu 22

gerases:
docker ps only lists containers on this node, so a stack with replicas
elsewhere looked missing and was redeployed every agent run. exists? now
compares compose services to `docker stack services`, strips a Swarm
digest pin from the image name, and treats a failed stack query as
absent. Compose files are loaded with Puppet::Util::Yaml.safe_load and
version 3.8 is accepted.

Fixes #968
Fixes #973

## Summary
Provide a detailed description of all the changes present in this pull request.

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.


gerases:
@skyamgarp, this fixes the docker_stack provider, would appreciate a review and if possible a merge.

jst-cyr:
@gerases: It seems that there are some syntax issues found by the Rubocop testing:

- [Spec Test Failure](https://github.com/puppetlabs/puppetlabs-docker/actions/runs/34644280253/job/104097695616?pr=1072)

```
Error: Style/BlockDelimiters: Prefer `do...end` for multi-line blocks without chaining. ([https://rubystyle.guide#single-line-blocks](https://rubystyle.guide/#single-line-blocks))
Error: Style/RegexpLiteral: Use `%r` around regular expression. ([https://rubystyle.guide#percent-r](https://rubystyle.guide/#percent-r))
```

gerases:
@jst-cyr, let me know if anything else is needed. Thank you!

jst-cyr:
@gerases : Thank you! Spec tests are passing, so I'll kick off some acceptance tests.

jst-cyr:
Running CI tests for this PR here: https://github.com/puppetlabs/puppetlabs-docker/pull/1073

jst-cyr:
**Acceptance Test Results:** https://github.com/puppetlabs/puppetlabs-docker/actions/runs/34889231355?pr=1073

Mostly green, but failures on CentOS-7, CentOS-8, Debian-10, Debian-11, and Ubuntu-20.04. I have compared these failures and they match identically to the failures against main that we are seeing in the nightly acceptance tests. This PR does not seem to be introducing new failures.

This still requires an engineering review, but testing seems good!

